PR #93, opened 2018-10-30
Polish preview app

JeroenDerks:


dionhaefner:
I'll just fix the types and do some cleanup and we're done :) Thanks for the good work.

PR #122, opened 2019-01-11
fix #119

### LINKED ISSUE #119: Explicit Colour Map Returning Transparent Tiles
Any request with an explicit colour map returns only transparent tiles. It doesn't seem to matter whether we use rgb tuples, rgba tuples or hex strings.

The issue appeared when upgrading from 0.4 to the newest version  (0.5.1.dev3+geed2b0e).
Example request:

`https://example.com/singleband/key1/key2/preview.png?colormap=explicit&explicit_color_map=%7B%220%22%3A%20%22ff0000ff%22%2C%20%221%22%3A%20%22ffffbeff%22%2C%20%222%22%3A%20%22e1e1e1ff%22%7D`

The URI encoded JSON is:
`{"0": "ff0000ff", "1": "ffffbeff", "2": "e1e1e1ff"}`

dionhaefner:
The transparency array had the wrong dtype (has to be uint8 to satisfy PIL).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ setup(
     ],
     extras_require={
         'test': [
-            'pytest>=3.5',
+            'pytest==4.0.2',
             'pytest-cov',
             'pytest-mypy',
             'pytest-flake8',
--- a/setup.py
+++ b/setup.py
@@ -77,7 +77,7 @@ setup(
             'pytest-cov',
             'pytest-mypy',
             'pytest-flake8',
-            'pytest-benchmark',
+            'pytest-benchmark==3.1.1',
             'attrs>=17.4.0',
             'codecov',
             'colorlog',
--- a/setup.py
+++ b/setup.py
@@ -77,7 +77,7 @@ setup(
             'pytest-cov',
             'pytest-mypy',
             'pytest-flake8',
-            'pytest-benchmark==3.1.1',
+            'pytest-benchmark',
             'attrs>=17.4.0',
             'codecov',
             'colorlog',
--- a/terracotta/image.py
+++ b/terracotta/image.py
@@ -79,11 +79,16 @@ def array_to_png(img_data: Array,
                     rgb.flatten(),
                     np.zeros(3 * (256 - len(colormap) - 1), dtype='uint8')
                 ))
-                transparency = np.concatenate((
+
+                # PIL expects paletted transparency as raw bytes
+                transparency_arr = np.concatenate((
                     np.zeros(1, dtype='uint8'),
                     alpha,
                     np.zeros(256 - len(colormap) - 1, dtype='uint8')
-                )).tobytes()
+                ))
+                assert transparency_arr.shape == (256,)
+                assert transparency_arr.dtype == 'uint8'
+                transparency = transparency_arr.tobytes()
 
             assert palette.shape == (3 * 256,), palette.shape
     else:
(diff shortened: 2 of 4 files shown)


dionhaefner:
We should also check how this wasn't caught by the tests.

mrpgraae:
The problem is that `test_singleband.py::test_singleband_explicit_colormap` which is supposed to test this, loads a tile that doesn't have any values overlapping the values given in the colormap. So this test always runs on an array of only 0's, which will always pass the asserts.

dionhaefner:
Ah, thought as much. Maybe we should just loop over all `np.unique` values in the loaded tile (I think we already do this somewhere?).

mrpgraae:
> 
> 
> Ah, thought as much. Maybe we should just loop over all `np.unique` values in the loaded tile (I think we already do this somewhere?).

So use the unique values to populate the colormap?

mrpgraae:
I scrambled the test raster to spread the values out a bit more, which caused the test to now fail (as it's supposed to) on my machine, but Travis is throwing up INTERNALERROR all over the place now. Don't think it's related to this.

mrpgraae:
>My proposal is to request the tile data first, extract the unique values, then build the explicit colormap based on some of those to ensure they actually occur in the tile.

Yes, I'll try this. Also seems like `pytest-benchmark` is causing the tests to error again.

dionhaefner:
I think we need to pin pytest, too (`pytest.warn` is deprecated IIRC, so they probably removed it)

mrpgraae:
> 
> 
> I think we need to pin pytest, too (`pytest.warn` is deprecated IIRC, so they probably removed it)

Yup https://docs.pytest.org/en/latest/changelog.html#pytest-4-1-0-2019-01-05
I'll pin it.

mrpgraae:
Actually we will still have the problem that the loaded tile doesn't contain any nodata, so I think I'll just make a new `categorical_raster` or so, with values in `[0, 255]`, `dtype == np.uint8` and every third pixel being nodata. That way it'll work with whatever colormap values.

mrpgraae:
Now the test works for the nodata and value consistency checks. However, we are now failing the check that checks that values not in the cmap should be transparent. I.e. we have values not in the cmap, that are not transparent in the final image.

Here are the values that fail the check
nocmap_fail

Here is the raw tile data for reference:
test_raster

mrpgraae:
Found the error. Just another dtype error.

mrpgraae:
I did however find what I would consider a bug in Numpy.
The failing line of code was this:
```python
np.all(img_data[~np.isin(tile_data, colormap.keys()), -1] == 0)
```
Which I think uses `np.asarray()` behind the scenes, to convert `colormap.keys()` to an ndarray.
The problem is that `dict.keys()` does not return an iterator or a list as one would expect, but some kind of special `dict_keys` object, which Numpy does not seem to handle well. Consider this strange behavior:

```
In [1]: import numpy as np

In [2]: int_dict = {1: 'a', 2: 'b', 3: 'c'}

In [3]: int_keys = int_dict.keys()

In [4]: int_keys
Out[4]: dict_keys([1, 2, 3])

In [5]: key_arr = np.asarray(int_keys)

In [6]: key_arr
Out[6]: array(dict_keys([1, 2, 3]), dtype=object)

In [7]: key_arr[0]
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-7-005c306646e2> in <module>
----> 1 key_arr[0]

IndexError: too many indices for array
```

So Numpy considers `dict_keys` an object and makes an object array, which might not be a bug, but sure is highly unexpected behavior. Furthermore, the IndexError is super weird, but I don't have much experience with object arrays in Numpy, so I'm not sure if that is expected behavior. I'll make an issue on Numpy in any case.

Issue #20, opened 2018-07-09
Support band math


dionhaefner:

pierotofy:
This would be amazing!

dionhaefner:
Basic functionality shouldn't be too hard, the real challenge is figuring out which operations to support, and how to pass them in the API. Happy for any ideas.

pierotofy:
Perhaps a subset of QGIS expressions could be a good list:

```
a + b      a plus b
a - b      a minus b
a * b      a multiplied by b
a / b      a divided by b
a % b      a modulo b (for example, 7 % 2 = 1, or 2 fits into 7 three
           times with remainder 1)
a ^ b      a power b (for example, 2^2=4 or 2^3=8)
a = b      a and b are equal
a > b      a is larger than b
a < b      a is smaller than b
a <> b     a and b are not equal
a != b     a and b are not equal
a <= b     a is less than or equal to b
a >= b     a is larger than or equal to b
+ a        positive sign
- a        negative value of a
a AND b      logical AND 
a OR b        logical OR
```

Then support for common math functions such as `cos`, `sin`, `acos`, `tan`, `sqrt`, etc.

https://docs.qgis.org/2.8/en/docs/user_manual/working_with_raster/raster_calculator.html

I perhaps a new endpoint such as:

```/calc/{keys}/{expression}/{tile_z}/{tile_x}/{tile_y}.png```

Or perhaps even:


Where expression is a URL-encoded string representation of the expression (referencing keys)

j08lue:
Thanks for pursuing this, @pierotofy. I think band math support would be a great addition. A few design questions:

1. Should `/calc` (or what we call it) always result in a single band or can you construct RGB (perhaps with 3 expressions)? -- I'd say RGB should be possible
2. Should an expression be a `&parameter` or a [`/<variable_section>`](http://flask.pocoo.org/docs/1.0/quickstart/#variable-rules) -- I would vote for the former for RGB support
3. FWIW: Sentinel Hub [uses full JS expressions](https://medium.com/sentinel-hub/ad-hoc-testing-of-algorithms-globally-8fb1f564f0f5) for arbitrarily complex band math. I don't think we should go there, though, should we?

dionhaefner:
We could support some functions that behave like NumPy ufuncs. The functionality you propose is `np.where`: `where(x > 50, 5, 10)`. I would be fine having `where` as a special function in the expressions, and it would be easy to implement since it's already in NumPy.

[...]

PR #128, opened 2019-01-28
Implement band math endpoint

dionhaefner:


dionhaefner:
@mrpgraae @j08lue @nmheim Could I ask one of you to take over from here? The feature is done from my perspective, but there are still some architectural decisions to be made (#20), the handler and flask route need test coverage, and the endpoint has to be documented.

terracotta/server/flask_api.py
```diff
@@ -110,6 +110,10 @@ def create_app(debug: bool = False, profile: bool = False) -> Flask:
 
     if profile:
         from werkzeug.contrib.profiler import ProfilerMiddleware
-        new_app.wsgi_app = ProfilerMiddleware(new_app.wsgi_app, restrictions=[30])
+        setattr(
+            new_app,
+            'wsgi_app',
+            ProfilerMiddleware(new_app.wsgi_app, restrictions=[30]
+        )
```
j08lue: ??

mrpgraae: I guess because this also works if `new_app.wsgi_app` doesn't exist already.

mrpgraae: No acutally, it won't. So I would also like to know 😄

dionhaefner: It's the recommended workaround for https://github.com/python/mypy/issues/2427

dionhaefner:
Done from my side. Needs more documentation, but so do all endpoints.

Issue #116, opened 2018-12-11
Compress cached tiles in memory


dionhaefner:
We currently run with 500MB of raster cache by default, which fills up very very quickly. We could alleviate this a bit by compressing the tiles before adding them to the cache (and decompress upon reading).

mrpgraae:
That is a really good idea IMO, I have had problems with a gunicorn + tc deployment eating up a lot of memory, due to each worker thread having it's own cache. So anything that optimizes caching would be a good idea I think. I would also expect the overhead of decompressing the tiles to be insignificant compared to network I/O.

dionhaefner:
Yea should be fairly low-hanging. Alternatively we could have another stab at making some driver thread-safe, then you could throw some threads in the mix. Third option could be to leverage an external cache like memcached or redis.

mrpgraae:
Actually, if we are going to start doing sessions for authentication as per #21, we might want to use something like Redis to store the sessions and then we might as well use it for caching too.

dionhaefner:
An external cache would be nice, but would be a lot of work, since we also need to support the corresponding AWS service. So we could do both, in-memory compression for now, and explore some of the more sophisticated options.

mrpgraae:
Yes, it should also be easy to implement the compression, so we should do that as a first go. For the more sophisticated options on Lambda, there's Amazon Elasticache https://aws.amazon.com/elasticache/ which seems to be just a Redis instance managed by AWS.

PR #133, opened 2019-03-27
Added data compression for cache

panakouris:
#116

### DIFF
--- a/terracotta/drivers/raster_base.py
+++ b/terracotta/drivers/raster_base.py
@@ -13,6 +13,8 @@ import operator
 import logging
 import math
 import warnings
+import zlib
+import sys
 
 import numpy as np
 from cachetools import cachedmethod, LFUCache
@@ -36,6 +38,42 @@ Number = TypeVar('Number', int, float)
 logger = logging.getLogger(__name__)
 
 
+class LFUCacheWithCompression(LFUCache):
+    """ Least-frequently-used Cache with data compression
+    """
+
+    def __getitem__(self, key: Any) -> np.ma.MaskedArray:
+        compressed_item = super().__getitem__(key)
+        return self._decompress_masked_array(compressed_item)
+
+    def __setitem__(self, key: Any, value: np.ma.MaskedArray) -> None:
+        super().__setitem__(key, self._compress_masked_array(value))
+
+    def _compress_masked_array(self, array: np.ma.MaskedArray) -> Tuple:
+        compressed_data = zlib.compress(array.data)
+        mask_to_int = np.packbits(array.mask.astype(np.uint8))
+        compressed_mask = zlib.compress(mask_to_int)
+        return (compressed_data,
+                compressed_mask,
+                array.dtype.name,
+                array.shape
+                )
+
+    def _decompress_masked_array(self, data: Tuple) -> np.ma.MaskedArray:
+        dt = np.dtype(data[2])
+        shape = data[3]
+        decompressed_data = np.frombuffer(zlib.decompress(data[0]), dtype=dt).reshape(shape)
+        decompressed_mask_packed = np.frombuffer(zlib.decompress(data[1]), dtype=np.uint8)
+        decompressed_mask_unpacked = np.unpackbits(decompressed_mask_packed)[:np.prod(shape)]
+        decompressed_mask = decompressed_mask_unpacked.reshape(shape)
+        return np.ma.masked_array(decompressed_data, mask=decompressed_mask)
+
+
+def _get_size_of(x: Tuple) -> int:
+    sizes = map(sys.getsizeof, x)
+    return sum(sizes)
+
+
 class RasterDriver(Driver):
     """Mixin that implements methods to load raster data from disk.
 
@@ -48,9 +86,9 @@ class RasterDriver(Driver):
     @abstractmethod
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         settings = get_settings()
-        self._raster_cache = LFUCache(
+        self._raster_cache = LFUCacheWithCompression(
             settings.RASTER_CACHE_SIZE,
-            getsizeof=operator.attrgetter('nbytes')
+            getsizeof=_get_size_of
         )
         self._executor = concurrent.futures.ThreadPoolExecutor(max_workers=3)
         super().__init__(*args, **kwargs)
--- a/terracotta/drivers/raster_base.py
+++ b/terracotta/drivers/raster_base.py
@@ -56,7 +56,7 @@ class LFUCacheWithCompression(LFUCache):
 
     def _compress_ma(self,
                      arr: np.ma.MaskedArray) -> CompressionTuple:
-        compressed_data = zlib.compress(arr.data)
+        compressed_data = zlib.compress(arr.data, self.compression_level)
         mask_to_int = np.packbits(arr.mask.astype(np.uint8))
         compressed_mask = zlib.compress(mask_to_int, self.compression_level)
         return (compressed_data,
--- a/terracotta/drivers/raster_base.py
+++ b/terracotta/drivers/raster_base.py
@@ -44,29 +44,36 @@ class LFUCacheWithCompression(LFUCache):
 
     def __getitem__(self, key: Any) -> np.ma.MaskedArray:
         compressed_item = super().__getitem__(key)
-        return self._decompress_masked_array(compressed_item)
+        return self._decompress_tuple(compressed_item)
 
     def __setitem__(self, key: Any, value: np.ma.MaskedArray) -> None:
-        super().__setitem__(key, self._compress_masked_array(value))
+        super().__setitem__(key, self._compress_ma(value))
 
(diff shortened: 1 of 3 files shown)


terracotta/drivers/raster_base.py
```diff
@@ -36,6 +38,42 @@
...
+        mask_to_int = np.packbits(array.mask.astype(np.uint8))
+        compressed_mask = zlib.compress(mask_to_int)
+        return (compressed_data,
+                compressed_mask,
+                array.dtype.name,
+                array.shape
+                )
+
+    def _decompress_masked_array(self, data: Tuple) -> np.ma.MaskedArray:
+        dt = np.dtype(data[2])
```
mrpgraae: Here you should probably do tuple unpacking, like:
```python
data_arr, mask_arr, dt, shape = data
```

terracotta/drivers/raster_base.py
```diff
@@ -36,6 +38,42 @@
...
+        shape = data[3]
+        decompressed_data = np.frombuffer(zlib.decompress(data[0]), dtype=dt).reshape(shape)
+        decompressed_mask_packed = np.frombuffer(zlib.decompress(data[1]), dtype=np.uint8)
+        decompressed_mask_unpacked = np.unpackbits(decompressed_mask_packed)[:np.prod(shape)]
+        decompressed_mask = decompressed_mask_unpacked.reshape(shape)
```
mrpgraae: Not super important, but I think it would be cleaner to re-use the same variable for decompression of the mask:
```python
decompressed_mask = np.frombuffer(zlib.decompress(data[1]), dtype=np.uint8)
decompressed_mask = np.unpackbits(decompressed_mask)[:np.prod(shape)]
decompressed_mask = decompressed_mask.reshape(shape)
```
This also has the benefit of re-using memory.

terracotta/drivers/raster_base.py
```diff
@@ -36,6 +38,42 @@
...
+        return np.ma.masked_array(decompressed_data, mask=decompressed_mask)
+
+
+def _get_size_of(x: Tuple) -> int:
+    sizes = map(sys.getsizeof, x)
```
dionhaefner: I have found `sys.getsizeof` to be blatantly wrong for NumPy arrays (which is why I used `nbytes` instead). Just a heads-up.

terracotta/drivers/raster_base.py
mrpgraae: maybe even just call it `mask`

terracotta/drivers/raster_base.py
```diff
@@ -36,6 +38,42 @@
...
+    """
+
+    def __getitem__(self, key: Any) -> np.ma.MaskedArray:
+        compressed_item = super().__getitem__(key)
+        return self._decompress_masked_array(compressed_item)
+
+    def __setitem__(self, key: Any, value: np.ma.MaskedArray) -> None:
+        super().__setitem__(key, self._compress_masked_array(value))
+
+    def _compress_masked_array(self, array: np.ma.MaskedArray) -> Tuple:
```
dionhaefner: Please add the full return type.

mrpgraae: The return type here should be `Tuple[np.ndarray, np.ndarray, str, Tuple[int, int]]`

mrpgraae:
Whoops, didn't mean to approve 😄

terracotta/drivers/raster_base.py
mrpgraae: Interesting. I guess we could change this to just add the `nbytes` together, the size of the dtype str and shape should be insignificant.

dionhaefner: That sounds more foolproof to me.

terracotta/drivers/raster_base.py
dionhaefner: I think it might be cleaner and more expressive to define a `CompressedArray` type (and set it to this value).

terracotta/drivers/raster_base.py
dionhaefner: Would even deserve a unit test to prevent this from breaking in the future. We have had problems with the cache not working as intended because of this.

terracotta/drivers/raster_base.py
mrpgraae: I agree, maybe a NamedTuple.

dionhaefner: In one-off cases like this I usually don't bother with custom data types, but we can leave it to @panakouris :wink:

terracotta/drivers/raster_base.py
panakouris: I used `sys.getsizeof` because we want to find the size of two compressed objects and not numpy arrays. I tested and seems to works fine on compressed objects, whereas in arrays or dictionaries it doesn't.

dionhaefner: Since you already checked it, could you please add a unit test? :) Just a simple sanity check, doesn't need to be accurate to the byte.

mrpgraae: Yes, seems we forgot that the compressed array are not numpy arrays 😄
They are probably some kind of Bytes object, right? So if `sys.getsizeof` should not work for whatever reason, a simple `len()` might do it.

tests/drivers/test_raster_drivers.py
```diff
@@ -4,6 +4,7 @@
 import rasterio.features
 from shapely.geometry import shape, MultiPolygon
 import numpy as np
+import zlib
```
dionhaefner: `zlib` is standard library, so it should come before these imports

tests/drivers/test_raster_drivers.py
```diff
@@ -562,3 +563,13 @@ def test_compute_metadata_unoptimized(unoptimized_raster_file):
...
     assert geometry_mismatch(shape(mtd['convex_hull']), convex_hull) < 1e-8
+
+
+def test_sizeof():
+    from terracotta.drivers.raster_base import _get_size_of
+    tile_shape = (256, 256)
+    data = zlib.compress(np.ones(tile_shape), 9)
+    mask = zlib.compress(np.zeros(tile_shape), 9)
+    size = _get_size_of((data, mask, 'float64', tile_shape))
+    assert size < 1550
```
dionhaefner:
```python
assert 1450 < size < 1550  # python 3 is neat
```

terracotta/drivers/raster_base.py
```diff
@@ -36,6 +38,49 @@
...
+
+    def __getitem__(self, key: Any) -> np.ma.MaskedArray:
+        compressed_item = super().__getitem__(key)
+        return self._decompress_tuple(compressed_item)
+
+    def __setitem__(self, key: Any, value: np.ma.MaskedArray) -> None:
+        super().__setitem__(key, self._compress_ma(value))
+
+    def _compress_ma(self,
+                     arr: np.ma.MaskedArray) -> Tuple[bytes,
```
dionhaefner: This hurts my sense of esthetics (and we don't do this anywhere else in the Terracotta codebase), so I think adding

```python
CompressedArray = Tuple[bytes, bytes, str, Tuple[int, int]]
```

somewhere in the header would be nicer and more expressive as to what you are doing.

terracotta/drivers/raster_base.py
```diff
@@ -36,6 +38,49 @@
...
+                                                      bytes,
+                                                      str,
+                                                      Tuple[int, int]]:
+        compressed_data = zlib.compress(arr.data)
+        mask_to_int = np.packbits(arr.mask.astype(np.uint8))
+        compressed_mask = zlib.compress(mask_to_int, 9)
```
dionhaefner: Can we expose the compression level as an option?

mrpgraae: Should just be implemented in: https://github.com/DHI-GRAS/terracotta/blob/master/terracotta/config.py

An then passed in to `LFUCacheWithCompression` in `RasterDriver.__init__`

panakouris: We can definately do it, but is there a need? As we discussed with @mrpgraae, we need to perform some benchmarks, and come up with the optimal compression level.

mrpgraae: Finding a good compression level would be good for setting the default value. Having it exposed to the user is nice, because then they can choose themselves what the memory / cpu-time tradeoff should be, if something else works better for them.

Also, as we talked about, I would assume the compute overhead of the compression to be insignificant compared to the total response time, so setting a high compression level as default is probably fine.

terracotta/drivers/raster_base.py
```diff
@@ -54,7 +58,7 @@ def _compress_ma(self,
                      arr: np.ma.MaskedArray) -> CompressionTuple:
         compressed_data = zlib.compress(arr.data)
```
dionhaefner: Wait, you are not using the compression level here. Was this on purpose?

panakouris: Good catch! I missed it, thanks

panakouris:
Closes #116

PR #154, opened 2019-10-30
Experimental: multiprocessing version of raster retrieval

dionhaefner:
GDAL has some problems with multi-threaded access to rasters: https://github.com/OSGeo/gdal/issues/1244

This is a multi-processing version of tile retrieval. A process pool is created the first time it is used (to keep startup time reasonable). Cache and database access is managed by the main process.

### DIFF
--- a/terracotta/drivers/raster_base.py
+++ b/terracotta/drivers/raster_base.py
@@ -6,16 +6,16 @@ Base class for drivers operating on physical raster files.
 from typing import (Any, Union, Mapping, Sequence, Dict, List, Tuple,
                     TypeVar, Optional, cast, TYPE_CHECKING)
 from abc import abstractmethod
-import concurrent.futures
+from concurrent.futures import Future
+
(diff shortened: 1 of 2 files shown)


terracotta/drivers/raster_base.py
```diff
@@ -36,7 +36,13 @@
...
 
 logger = logging.getLogger(__name__)
-executor = ProcessPoolExecutor(max_workers=3)
+
+try:
+    # this fails on architectures without /dev/shm
+    executor = ProcessPoolExecutor(max_workers=3)
+except OSError:
+    # fall back to serial evaluation
+    executor = ThreadPoolExecutor(max_workers=1)
```
j08lue: nice!

vincentsarago:
@dionhaefner did you run any benchmarking to see any performance improvement using multiprocess instead of multithreading ? I wonder how it will perform within AWS Lambda ?

PR #158, opened 2019-11-26
Explicitly compute overview levels

dionhaefner:
Based on https://github.com/cogeotiff/rio-tiler/pull/132

Overall some performance improvements, but we will have to test more to see whether it is worth it.


dionhaefner:
No benefit with the new default transform.

PR #166, opened 2020-02-28
make CORS allowed origins configurable

### LINKED ISSUE #160: CORS for map tiles?
We currently enable CORS only for `metadata_api`, not for `tile_api` or `spec_api`:

https://github.com/DHI-GRAS/terracotta/blob/3c9feff9910382da1c827f11d8a84e7f08a44970/terracotta/server/flask_api.py#L21

However, we need CORS also for map tiles when using Mapbox GL as a client, because it fetches the data inside the web client, rather than just letting the browser load the images directly, like Leaflet does it.

What is the rationale behind not enabling CORS for all endpoints? And if we were to change that, should we just enable CORS everywhere or make it a configurable option (`CORS=false` or so)?

j08lue:
Make CORS allowed origins configurable for metadata and tiles via new environment variables:
* `TC_ALLOWED_ORIGINS_METADATA = '["*"]'`
* `TC_ALLOWED_ORIGINS_TILES = '[]'`

Where empty list `[]` means no allowed origins.

The changes are backward-compatible, so even when the user omits the new environment variables but updates Terracotta, the behavior will be the same as before: all origins allowed for `/metadata`, no cross-origin headers for tiles.

Closes #160


terracotta/config.py
```diff
@@ -105,6 +111,9 @@ class SettingSchema(Schema):
         validate=validate.OneOf(['nearest', 'linear', 'cubic', 'average'])
     )
 
+    ALLOWED_ORIGINS_METADATA = fields.String()
```
dionhaefner: Shouldn't these be more like list-of-string?

dionhaefner:
For testing this I suggest you add a test that sends a request to the server and inspects the returned headers (should contain `Access-Control-Allow-Origin` with the allowed origins).

j08lue:
Yes, doing just that. How about what I just committed? And any idea why the last test does not pass?

terracotta/config.py
j08lue: Could be, too. OK then.

dionhaefner: I think list of string is less ambiguous, so we should support that.

tests/server/test_cors.py
```diff
@@ -0,0 +1,78 @@
...
+    return '/metadata/val11/x/val12/'
+
+
+@pytest.fixture(scope='module')
+def valid_singleband_path(raster_file_xyz):
+    x, y, z = raster_file_xyz
+    return f'/singleband/val11/x/val12/{z}/{x}/{y}.png'
+
+
+def test_cors_defaults(
```
dionhaefner: In Terracotta, we break lines in function headers like this:

```python
def func(val1, val2,
              val3, val4):
```

tests/server/test_cors.py
```diff
@@ -0,0 +1,78 @@
...
+    client_defaults, use_testdb, valid_metadata_path, valid_singleband_path
+):
+    rv = client_defaults.get(valid_metadata_path, headers=HEADERS_STRANGE_ORIGIN)
```
dionhaefner: What are these headers for?

terracotta/server/flask_api.py
```diff
@@ -89,6 +80,19 @@ def create_app(debug: bool = False, profile: bool = False) -> Flask:
...
+    new_app = Flask('terracotta.server')
+    new_app.debug = debug
+
+    # suppress implicit sort of JSON responses
+    new_app.config['JSON_SORT_KEYS'] = False
+
+    # CORS
+    settings = get_settings()
+    if settings.ALLOWED_ORIGINS_METADATA:
+        CORS(metadata_api, origins=settings.ALLOWED_ORIGINS_METADATA)
```
dionhaefner: It looks like this modifies the (global) blueprints in-place, so resetting the setting doesn't revert this. It might be cleanest to copy the global blueprints in this function if that's possible.

dionhaefner: On second thought, having multiple instances of Terracotta in the same Python thread is undefined behavior anyway. So you can also just set CORS unconditionally (with whatever is set in the settings, including no allowed origins). You might have to set the fixture scope for the test client to local in that case, so pytest creates a new one for every test.

dionhaefner: Maybe add a comment to the docs that the parameter has to be set before creating the app.

tests/server/test_cors.py
j08lue: I actually do not need them. The hostname at runtime work just as well. Will remove.

terracotta/server/flask_api.py
j08lue: > Maybe add a comment to the docs that the parameter has to be set before creating the app.

Oooh, yes, I did not think of that. Users must definitely be (made) aware that some config changes only go live when the TC server restarts, which in the case of Lambda happens when the functions goes cold or is updated, I guess.

j08lue: Since these Flask apps, blueprints and stuff have so much global state (and also our `terracotta.settings`), maybe we should ensure that tests run in separate processes with [`pytest-xdist` and `--forked`](https://stackoverflow.com/a/48310939/512111). 

I could make my tests pass by running disabled CORS first and enabled after, but that this works is just coincidental.

dionhaefner:
A bit hacky with the deepcopy, but seems to work.

I think we should try our best to minimize global state (except our settings, which are easy enough to reset) before giving up and using separate processes.

j08lue:
> I think we should try our best to minimize global state (except our settings, which are easy enough to reset) before giving up and using separate processes.

Very much prefer that, too.

> A bit hacky with the deepcopy, but seems to work.

Since blueprints do not have a copy method, this will have to do.

dionhaefner:
Only thing missing is support for list of string (or documentation on how to allow multiple origins), then we're good to go.

j08lue:
> Only thing missing is support for list of string (or documentation on how to allow multiple origins), then we're good to go.

Yeah, I tried making that work - parsing that from an environment variable with `marshmallow`. Did not get too far and it seemed a bit messy. But I'll make it work, if you prefer that to just str.

dionhaefner:
We can leave it as a string if the alternative is too messy, but then we need to tell the user somehow how to whitelist several origins.

dionhaefner:
Supporting only list of string wouldn't be too bad, either.

terracotta/server/flask_api.py
j08lue: Avoiding global state is nicer, so no need for xdist.

j08lue:
> Supporting only list of string wouldn't be too bad, either.

I implemented this now, re-using the JSON value support we already had. So you need to set `ALLOWED_ORIGINS_TILES = '["*"]'` for all origins now. Not pretty, but supports multiple values without much extra code. I suggest we leave it at that, just add a note in the docs.

dionhaefner:
Looks good to me.

j08lue:
@dionhaefner please review and merge if ready

j08lue:
image

PR #167, opened 2020-02-29
run pytest with xdist and forked

j08lue:
As discussed in the context of another PR (https://github.com/DHI-GRAS/terracotta/pull/166#pullrequestreview-366467067), Flask apps like Terracotta have a lot of global state, since the fully configured app is instantiated upon import:

https://github.com/DHI-GRAS/terracotta/blob/279b840cd5d8372b5895beba6c71498656fb468b/terracotta/server/app.py#L16-L19

Multiple instances of the app are not meant to run in the same kernel, but this is what happens when pytest executes the tests one after the other in the same kernel and the state of one instance can influence the outcome of the other.

Maybe we can change that with `pytest-xdist`, where each test runs in its own process, adding considerable overhead, but perhaps more security.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ before_script:
   - pip freeze
 
 script:
-  - python -m pytest --color=yes --cov=terracotta --mysql-server="root@127.0.0.1"
+  - python -m pytest --color=yes --cov=terracotta --forked --mysql-server="root@127.0.0.1"
   - python -m pytest --color=yes tests/benchmarks.py
 
 after_success:
--- a/setup.py
+++ b/setup.py
@@ -82,6 +82,7 @@ setup(
             'pytest-mypy',
             'pytest-flake8',
             'pytest-benchmark',
+            'pytest-xdist',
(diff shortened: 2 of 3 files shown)


j08lue:
Y tho? https://travis-ci.com/DHI-GRAS/terracotta/jobs/292591107#L1544

mrpgraae:
>Y tho?

ExpressionParser uses a recursive implementation (see https://github.com/DHI-GRAS/terracotta/blob/2ac1b4e48e3571ad5e95a3436c376b35ccd372fe/terracotta/expressions.py and https://docs.python.org/3/library/ast.html#ast.NodeVisitor.visit).

Parsing expressions is also inherently a recursive problem and any non-recursive implementation is likely to be nasty I think, so it makes sense to implement it like so.

You could increase the max recursion depth in the test
```python
import sys
sys.setrecursionlimit(BIG_NUMBER)
```

dionhaefner:
I think we agree that `xdist` is not needed at this point.

PR #168, opened 2020-03-01
Update wsgi.rst

YuhangCh:
Modify the sock position and application entry to correspond to the context


dionhaefner:
Thank you for the contribution. Could you give some context with that? Why is it preferable to have the socket in the same location as the data?

YuhangCh:
According to the explanation of the service file:
- Binding to a unix socket file terracotta.sock in the working directory

dionhaefner:
We also have

```
WorkingDirectory=/mnt/data
```

so the socket file should already go there. Did you find that to not be the case?

YuhangCh:
No, the socket file is already in the working directory. Adding an absolute path seems redundant.I removed it.

dionhaefner:
Looks good. Thanks!

PR #169, opened 2020-03-25
Switch back to GDAL's default transform

dionhaefner:
See https://github.com/cogeotiff/rio-tiler/issues/164

### DIFF
--- a/terracotta/drivers/raster_base.py
+++ b/terracotta/drivers/raster_base.py
@@ -11,13 +11,11 @@ from concurrent.futures import Future, Executor, ProcessPoolExecutor, ThreadPool
 import contextlib
 import functools
 import logging
-import math
 import warnings
 import threading
 
 import numpy as np
 import cachetools
-from affine import Affine
 
 if TYPE_CHECKING:  # pragma: no cover
     from rasterio.io import DatasetReader  # noqa: F401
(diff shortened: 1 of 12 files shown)


terracotta/drivers/raster_base.py
```diff
@@ -479,24 +429,43 @@ def _get_raster_tile(cls, path: str, *,
...
 
-            if bounds is None:
-                bounds = dst_bounds
+            # make sure VRT resolves the entire tile
+            tile_transform = transform.from_bounds(*tile_bounds, *tile_size)
+            tile_res = (abs(tile_transform.a), abs(tile_transform.e))
+
+            if any(tile_res < dst_res):
+                dst_res = tile_res
+                resampling_enum = cls._get_resampling_enum('nearest')
```
vincentsarago: @dionhaefner this is to make sure you are no oversampling right ?

dionhaefner: In that case, the VRT already has the exact shape and resolution required for the tile, so any further resampling is unnecessary. 

(assuming you mean setting resampling to nearest)

dionhaefner: To be clear, this is a minor performance optimization. Results should be identical.

PR #188, opened 2020-10-08
Driver caching & MySQL credentials

### LINKED ISSUE #187: Can't connect to MySQL server if password contains '#' character
`terracotta.get_driver(mysql_path)`

Fails when _mysql_path_ contains '#' in any part. It's actually because of  `urllib.parse` , which fails to parse the URL components if any of the component has '#' in it. There may be other special characters too causing the issue. The password part of the URL has high probability of having these characters. My example : 
```
mysql_path = 'mysql://root:admin#!@123@localhost:3306/db'
```
_Also if you move the '#' around in the password itself, the parsing result is different and hence the errors too_

I solved it by using `quote/unquote` of `urllib.parse` : 
`mysql_path = 'mysql://root:'+ quote('admin#!@123') + '@localhost:3306/db'`

And changed 
https://github.com/DHI-GRAS/terracotta/blob/5776df9a58cd47b4bf0fcb29070ff40ecdf41d98/terracotta/drivers/mysql.py#L118

`password=urlparse.unquote(con_params.password) if con_params.password else None,`

Is this solution right way to handle the issue?

dionhaefner:
- Use normalized paths for driver cache (so `https://example.com`, `example.com/`, ~`https://example.com/foo/../`~, `root:pw@example.com` all return the same driver)
- Introduce config vars for MySQL credentials (fixes #187)

Needs some more test coverage.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,6 @@ setup(
         'pyyaml>=3.10',  # downstream dependency of apispec
         'shapely',
         'rasterio>=1.0',
-        'url_normalize',
         'shapely',
         'toml',
         'tqdm'
--- a/setup.py
+++ b/setup.py
@@ -69,6 +69,7 @@ setup(
         'pyyaml>=3.10',  # downstream dependency of apispec
         'shapely',
         'rasterio>=1.0',
+        'url_normalize',
         'shapely',
         'toml',
         'tqdm'
--- a/setup.py
+++ b/setup.py
@@ -88,18 +88,18 @@ setup(
             'flake8',
             'matplotlib',
             'moto',
-            'pymysql'
+            'pymysql<0.10'
         ],
         'docs': [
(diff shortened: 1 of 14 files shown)


j08lue:
I originally had a much simpler solution to #187 in mind: just replacing `urlparse` in `auto_detect_provider` by some simple string heuristics (`.startswith` etc) to to find the driver scheme, since that seemed to be all we were using `urlparse` for.

https://github.com/DHI-GRAS/terracotta/blob/86bdac31078e72fbe2ecaf08a61f15f9429389a0/terracotta/drivers/__init__.py#L31-L41

This is much more robust, which is nice.

dionhaefner:
Aha, not quite - the MySQL driver uses urlparse to extract path and credentials from the driver path, too. So we need urlparse regardless.

PR #192, opened 2021-02-24
Add option to skip existing datasets on ingest

### LINKED ISSUE #191: Add option to skip existing datasets when ingesting
This picks up from the discussion in https://github.com/DHI-GRAS/terracotta/issues/190#issuecomment-785040789.

You can already get the existing datasets (keys and paths) with 

```python
import terracotta
driver = terracotta.get_driver(DB_PATH)
existing_datasets = driver.get_datasets()
```

So it is just about adding a line somewhere below
https://github.com/DHI-GRAS/terracotta/blob/018c4f7a059086e0eee6f923370b7a03cd8ff70a/terracotta/scripts/ingest.py#L69-L71 
that filters out the existing datasets, if a CLI flag is set.

I can fix this one, unless you are already on it, @dionhaefner? 😉

j08lue:
- [x] Add CLI option to `terracotta ingest --skip-existing` datasets
- [x] Add tests
- [ ] See whether it works

Closes #191

### DIFF
--- a/terracotta/scripts/ingest.py
+++ b/terracotta/scripts/ingest.py
@@ -71,7 +71,8 @@ def ingest(raster_pattern: RasterPatternType,
 
     if not output_file.is_file():
         driver.create(keys)
-    elif skip_existing:
+
+    if skip_existing:
         existing = driver.get_datasets()
         raster_files = {key: path for key, path in raster_files.items() if key not in existing}
 
--- a/terracotta/scripts/ingest.py
+++ b/terracotta/scripts/ingest.py
@@ -25,12 +25,15 @@ logger = logging.getLogger(__name__)
                    '(will be computed on first request instead)')
 @click.option('--rgb-key', default=None,
               help='Key to use for RGB compositing [default: last key in pattern]')
+@click.option('--skip-existing', is_flag=True, default=False,
+              help='Skip existing datasets by key')
 @click.option('-q', '--quiet', is_flag=True, default=False, show_default=True,
               help='Suppress all output to stdout')
 def ingest(raster_pattern: RasterPatternType,
            output_file: Path,
            skip_metadata: bool = False,
            rgb_key: str = None,
+           skip_existing: bool = False,
            quiet: bool = False) -> None:
     """Ingest a collection of raster files into a (new or existing) SQLite database.
 
@@ -42,7 +45,7 @@ def ingest(raster_pattern: RasterPatternType,
 
     The empty group {} is replaced by a wildcard matching anything (similar to * in glob patterns).
 
-    Existing datasets are silently overwritten.
+    Existing datasets are silently overwritten, unless you set --skip-existing.
 
     This command only supports the creation of a simple, local SQLite database without any
     additional metadata. For more sophisticated use cases use the Terracotta Python API.
@@ -68,6 +71,9 @@ def ingest(raster_pattern: RasterPatternType,
 
     if not output_file.is_file():
         driver.create(keys)
+    elif skip_existing:
+        existing = driver.get_datasets()
+        raster_files = {key: path for key, path in raster_files.items() if key not in existing}
 
     if tuple(keys) != driver.key_names:
         click.echo(
--- a/tests/scripts/test_ingest.py
+++ b/tests/scripts/test_ingest.py
@@ -158,9 +158,11 @@ def test_reingest(skip_existing, raster_file, tmpworkdir):
     from terracotta import get_driver
 
     same_name = 'myimage'
-    infiles = [f'dir1/{same_name}.tif', f'dir2/{same_name}.tif', f'dir3/{same_name}.tif']
-    for infile in infiles:
-        temp_infile = tmpworkdir / infile
+    infiles = [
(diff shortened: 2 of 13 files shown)


j08lue:
Tons of tests failing because of this `PerformanceWarning`?

```
terracotta.exceptions.PerformanceWarning: Raster file /tmp/pytest-of-travis/pytest-1/raster0/img-nodata.tif is not a valid cloud-optimized GeoTIFF. Any interaction with it will be significantly slower. Consider optimizing it through `terracotta optimize-rasters` before ingestion.
```

https://travis-ci.com/github/DHI-GRAS/terracotta/jobs/486040485#L14501

dionhaefner:
I fixed most of the failing tests, no idea what's up with the unclosed socket one though. And mypy complaining about connect not being a context manager?

j08lue:
Awesome!

> And mypy complaining about connect not being a context manager?

That might be the old PyMySQL degradation that they fixed recently. We had pinned the version though, and it was a context manager then and should be now again... https://github.com/PyMySQL/PyMySQL/pull/886

dionhaefner:
Yikes, besides that we also have some hefty performance regressions... tile retrieval now taking about 3x as long as it used to. I guess this is GDAL3's work...

vincentsarago:
@dionhaefner 👋 https://github.com/cogeotiff/rio-tiler#gdal30--proj60-performances-issue

dionhaefner:
🙏 Hallelujah.

I pinned rasterio now, a factor of 3 in performance is no joke 😕

terracotta/scripts/ingest.py
```diff
@@ -68,6 +71,9 @@ def push_to_last(seq: Sequence[Any], index: int) -> Tuple[Any, ...]:
 
     if not output_file.is_file():
         driver.create(keys)
+    elif skip_existing:
```
dionhaefner, line 74: This could just be `if skip_existing`.

tests/scripts/test_ingest.py
```diff
@@ -152,6 +152,41 @@ def test_ingest_append(raster_file, tmpworkdir):
     assert all((ds,) in driver.get_datasets() for ds in ('img1', 'img2'))
 
 
+@pytest.mark.parametrize('addflags', [['--skip-existing', ], []])
```
dionhaefner, line 155: I think I would prefer simpler logic here. Something like a boolean `skip_existing` flag.

tests/scripts/test_ingest.py
```diff
@@ -152,6 +152,41 @@ def test_ingest_append(raster_file, tmpworkdir):
...
+def test_ingest_append_overwrite(addflags, raster_file, tmpworkdir):
```
dionhaefner, line 156: `test_reingest`?

tests/scripts/test_ingest.py
```diff
@@ -152,6 +152,41 @@ def test_ingest_append(raster_file, tmpworkdir):
...
+    assert outfile.check()
+
+    from terracotta import get_driver
+    driver = get_driver(str(outfile), provider='sqlite')
+    assert driver.key_names == ('name',)
+    keys = list(driver.get_datasets())
+    assert len(keys) == 1
+    assert len(keys[0]) == 1
+    assert keys[0] == (same_name, )
+
```
dionhaefner, line 188: This should test that the raster points to the correct path, too

tests/scripts/test_ingest.py
j08lue: It is a boolean flag. This parametrization just tests both - overwrite with same key when the flag is set and when it is not.

tests/scripts/test_ingest.py
j08lue: Yes.

tests/scripts/test_ingest.py
j08lue: Good point.

terracotta/scripts/ingest.py
j08lue: Why would you prefer that? For clarity? I guess there is no point in checking for existing datasets when the db has just been created?

tests/scripts/test_ingest.py
j08lue: I get it now.

tests/scripts/test_ingest.py
j08lue: Done.

tests/scripts/test_ingest.py
j08lue: Done.

tests/scripts/test_ingest.py
j08lue: Done.

terracotta/scripts/ingest.py
dionhaefner: It's just something I stumbled over when I read your changes. I had to stop and think why this would only apply in the `else` case. IMO it also improves maintainability; you could change or remove the first check without impacting the second.

dionhaefner: It's a nitpick, of course :)

Issue #182, opened 2020-07-20
Demo website isn't working 


plant99:
https://terracotta-python.readthedocs.io/en/latest/preview-app.html on loading, gives the following error in browser console
```
Uncaught (in promise) TypeError: Cannot read property 'join' of undefined
    at updateColormap (main.js:593)
    at initUI (main.js:344)
    at main.js:1287
```

dionhaefner:
Whoops! Thanks for letting us know.

The demo server runs an outdated version of Terracotta. I'm not sure how to best fix this (and avoid regressions in the future). Perhaps the Satlas and TC demo server should be separate deployments, where the latter one is deployed automatically on new releases (but they can use the same data). Would that be doable @j08lue?

j08lue:
> Perhaps the Satlas and TC demo server

Yes. I'll set up a dedicated demo deployment (AWS Lambda) with latest Terracotta.

How about we use some publicly available set of COGs and just index that? Any preferences? Maybe we can find some under this Twitter handle https://twitter.com/cogeotiff ...

dionhaefner:
Sure! Anything that's pretty to look at will do for me, really. If you can handle the AWS side of things and send me an access key, I can handle the Travis deployment. I think I would do it like this:

- Run on new releases, before the PyPI deployment
- Re-build database from scratch (so the rasters can't be too big)
- Bundle the SQLite database with the Lambda code (then we don't need a dedicated S3 bucket for data)

dionhaefner:
That hinges on the open data being cloud-optimized, though. Do you have any experience with that? Otherwise we would need to dump that into a dedicated S3 bucket (maybe the safer option anyway).

j08lue:
Yes, the data I suggested is all COG. But we could also use our own data. The best would probably be to have both float and categorical data, so that we can nicely demo the color stretch functionality and users can also explore custom categorical color maps etc. So maybe I should just upload some of the old anyways.

Regarding deployment/upgrades: you are not talking about doing this in CI/CD, are you? Because, AFAIK, Zappa still requires an admin-ish account to create all the resources and roles that the S3 / Lambda / API Gateway stack needs and I don't want to store that anywhere, not even in GitHub/Travis secrets.

dionhaefner:
> The best would probably be to have both float and categorical data, so that we can nicely demo the color stretch functionality and users can also explore custom categorical color maps etc.

Don't bother with categorical data, the viewer doesn't expose that functionality.

[...]

Issue #135, opened 2019-04-03
Mutli-dataset Selection in Terracotta connect


mccarthyryanc:
In the Terracotta Preview App, `terracotta connect`, you can click and select individual datasets in the table search results. Would it be possible to add `ctrl+click` and `shift+click` support in order to view multiple datasets at the same time?

mrpgraae:
That is certainly possible and would be a nice feature. It should also only require a change to the preview app itself. Each dataset would just be another leaflet layer.

dionhaefner:
As @mrpgraae mentions, this is not a fundamental issue at all. However, this would require quite substantial changes to the preview app:

- Contrast sliders would have to be remodeled (or should the same contrast apply to all layers?)
- Mechanism to move layers up / down the stack?
- RGB and singleband not mutually exclusive anymore?
- Keep layers active across searches

I think in this case we would like a list of active layers that can be manipulated by the user (drag new layers in, press to toggle, remove, reorder).

Since my Javascript fu is limited, we will have to rely on volunteers for this one (inside DHI GRAS or from the outside).

mrpgraae:
>Contrast sliders would have to be remodeled (or should the same contrast apply to all layers?)

Maybe as a first go it would be alright to have the same contrast applied to all layers. If you want to compare values from different layers, this is probably what you want anyways.

>Mechanism to move layers up / down the stack?

and 

>Keep layers active across searches

Would be nice, but maybe not strictly necessary for a first implementation.

>RGB and singleband not mutually exclusive anymore?

This is worse. Any ideas on what to do here would be appreciated.

mccarthyryanc:
@mrpgraae your suggestions would fit my use case! I have a simple dataset: non-overlapping layers, same datatype across layer, single band. Keeping the layers active across searches would be nice, but not critical for an initial implementation.

@dionhaefner My javascript fu is also limited, but I will take a look and see if I can get something working.

mrpgraae:
Sounds good @mccarthyryanc! I suffer from the same JS limitations as @dionhaefner, but I'll gladly help and give advice if needed.

Issue #91, opened 2018-10-26
Write a client library


dionhaefner:
Terracotta would only be a complete A-Z solution if we also provided a client library in Javascript to do some of the heavier lifting.

I don't see this happen anytime soon, since we would need someone who is way way better at Javascript than me.

robmarkcole:
Perhaps just write some docs for use with an existing solution e.g. https://terria.io/

dionhaefner:
Great point, thanks!

dionhaefner:
For the record, it's really easy to do on something like [geojson.io](https://geojson.io):

- Click `Meta -> Add Map Layer`
- Enter URL to dataset in the form `http://localhost:5000/singleband/key1/.../keyN/{z}/{x}/{y}.png` (with Terracotta serving on `localhost:5000` in this case; you can get the keys from `http://localhost:5000/datasets`)

Works similarly on everything that uses Leaflet.JS, via `map.addLayer`.

I didn't know terria.io, but it looks really interesting, so thanks for the pointer.

PR #194, opened 2021-04-14
Feature/react ts client preview

### LINKED ISSUE #140: Preview app does not support datasets where min == max
Activating the layer does not work. I get

```
Error: noUiSlider (12.0.0): 'range' 'min' and 'max' cannot be equal.
```

in the JS console. Seems like we will have to catch this case...

### LINKED ISSUE #195: Rewrite TC preview app
For a better UI, maintainability, and consistency, the TC preview app can be re-written in React and TS.

The interface can be similar to the apps we keep building such as [satlas.dk](satlas.dk) or [urban.satlas.dk](urban.satlas.dk).

bertearazvan:
This PR aims to rewrite the Terracotta preview app using React and Typescript based on the create-react-app boilerplate.
- it adds the webpack configuration for local development and production build.
- it adds the flask script to run the app
- it recreates the app using the same kind of UI as satlas.dk

Close #195 
Close #140


bertearazvan:
d6aeeff33e43970ba3e26227f36d7fa2

dionhaefner:
Are you planning to address some of our long-standing issues? #91 #135 #140 #182

bertearazvan:
Probably #140 is fixed(currently you can only slide, and not use an input for those min/max values - unsure if this is an issue. The precision of the slider for singleband values is `0.01`, not sure if it should be more precise than this) and I'll look into #182, the other two seem a bit complex for this PR. 

There are discussions going around #91 about having a client library that wraps TC with a Map provider - probably through deck.gl. Stay tuned

As about #135, hmm interesting feature, sparks my interest so I might look into it, but as a separate PR.

dionhaefner:
OK. I personally prefer entering values to sliding, but your call if you want to support that.

#140 was not about this, though. The bug arose when a raster had a constant value, e.g. all 0. Then, the slider would break. So you should probably test this case to make sure it's fixed.

terracotta/client/app/src/App.tsx
```diff
@@ -82,7 +82,7 @@ const App: FC<Props> = ({ hostnameProp }) => {
 
   const initializeApp = (hostname: string | undefined) => {
     // sanitize hostname
-    // hostname = 'https://4opg6b5hc3.execute-api.eu-central-1.amazonaws.com/development'
+    hostname = 'https://4opg6b5hc3.execute-api.eu-central-1.amazonaws.com/development'
```
j08lue, line 85: Is this only used during development? We should probably not include this in the code - or use an "official" demo deployment (which we currently are lacking).

terracotta/client/app/src/colormap/SinglebandSelector.tsx
```diff
@@ -0,0 +1,109 @@
...
+                        activeSinglebandRange !== undefined && 
+                        activeDataset !== undefined &&
+                        (
+                        <>
+                            <Slider 
+                                noNumbers 
+                                getValueCommitted={value => Array.isArray(value) && onSetRangeValue(value)} 
+                                getValue={(value: number | number[]) => Array.isArray(value) && setLocalRange(value)}
+                                defaultValue={activeSinglebandRange} 
+                                min={datasets[activeDataset - page * limit]?.range[0] || 0}
```
rihr, line 88: Could put in consts

terracotta/client/app/src/sidebar-datasets/DatasetPreview.tsx
```diff
@@ -0,0 +1,127 @@
...
+                                <code>
+                                    {'Metadata\n'}
+                                </code>
+                            </Typography>
+                                <code style={{ whiteSpace: 'pre' }}>
+                                    {'{\n'}
+                                    {Object.keys(dataset).reduce(
+                                        (acc: string[], keyItem: string, j: number) => 
+                                           {
+                                                const neededKeys = ['mean', 'range', 'stdev', 'valid_percentage'];
```
rihr, line 86: Could be a function var

terracotta/client/app/src/sidebar-datasets/SidebarDatasets.tsx
```diff
@@ -0,0 +1,318 @@
...
+
+    
+    }
+
+    // useEffect(() => {
+
+    //     void getKeys(host)
+
+    // }, [host]) // eslint-disable-line react-hooks/exhaustive-deps
+
```
rihr, line 152:
```suggestion

```

rihr (approved):

dionhaefner:
Thanks for the implementation, it sure looks more professional now.

I just tried it, and I do have some comments:

- The info text when you click on details seems superfluous now. Why not just remove it? It's hard to find anyway.
- The bold `/apidoc` in the details section looks weird to me.
- Raster listing starts at page 0, and no indication how many pages there are.
- No padding between metadata and preview image looks cramped.
- <s>What happened to the zoom to the dataset? Imaging you have a raster of one yard in New Zealand, how are you ever going to find it on the world map?</s> Just saw that it *does* zoom in, but not if you select the first dataset in the list :thinking:
- The function of the "change base map" button is not obvious before you click on it. Couldn't this be "satellite view" / "street view" with a matching icon instead?
- Sidebar looks very large on smaller / non-maximized windows, and is not resizable anymore:

Screen Shot 2021-04-26 at 10 11 13

- I just saw that you can minimize the sidebar. That arrow is hard to find. Couldn't this be on the left side of the sidebar?
- **RGB view is broken, there is no way to select different rasters for the bands.**
- I don't think the RGB contrast sliders work as they should. They just have a range of 0-255 instead of the range of the rasters.
- Key descriptions are not part of the UI anymore?
- Some more vertical breathing room would be nice.

No one has asked for my review, so you can do with this feedback whatever you want.

j08lue:
Thanks for the ex post facto review, @dionhaefner. 😄 Thought it was nice to push the remake out there, great you are already picking up the revision on another PR, @bertearazvan.

PR #200, opened 2021-04-26
Remove hostname url hardcoded

### LINKED ISSUE #201: Remove hardcoded hostname
In the App.tsx, the hostname url should not be hardcoded.

bertearazvan:
For development purposes, I have used that URL. When merging, that URL should not be hardcoded there.

Close #201


dionhaefner:
Aha, this needs a rebuild, too.

PR #203, opened 2021-04-26
TC preview fixes and enhancements

### LINKED ISSUE #199: Preview app bugs and enhancements
Thanks for the implementation, it sure looks more professional now.

I just tried it, and I do have some comments:

- The info text when you click on details seems superfluous now. Why not just remove it? It's hard to find anyway.
- The bold `/apidoc` in the details section looks weird to me.
- Raster listing starts at page 0, and no indication how many pages there are.
- No padding between metadata and preview image looks cramped.
- <s>What happened to the zoom to the dataset? Imaging you have a raster of one yard in New Zealand, how are you ever going to find it on the world map?</s> Just saw that it *does* zoom in, but not if you select the first dataset in the list :thinking:
- The function of the "change base map" button is not obvious before you click on it. Couldn't this be "satellite view" / "street view" with a matching icon instead?
- Sidebar looks very large on smaller / non-maximized windows, and is not resizable anymore:

Screen Shot 2021-04-26 at 10 11 13

- I just saw that you can minimize the sidebar. That arrow is hard to find. Couldn't this be on the left side of the sidebar?
- **RGB view is broken, there is no way to select different rasters for the bands.**
- I don't think the RGB contrast sliders work as they should. They just have a range of 0-255 instead of the range of the rasters.
- Key descriptions are not part of the UI anymore?
- Some more vertical breathing room would be nice.

No one has asked for my review, so you can do with this feedback whatever you want.

_Originally posted by @dionhaefner in https://github.com/DHI-GRAS/terracotta/issues/194#issuecomment-826627029_

bertearazvan:
Closes #199 

- [x] Fix bug with zoom on the first dataset.
- [x] Change icon for basemap toggle
- [x] Add default rgb values from metadata
- [x] Add link to metadata for that dataset
- [x] Remove description
- [x] Remove bold from `/apidoc`
- [x] Add resizer for sidebar
- [x] Add description for keys
- [x] Update readme with screenshot
~~- [ ] move sidebar to the ~~right~~ left~~ As our apps all have the sidebar to the right and the UX department said that the users did not encounter problems/didn't have a negative feedback on having the sidebar to the right, for now we will keep it on the right side of the screen for consistency across our apps.


bertearazvan:
If the keys have a description, they can be found by hovering that key as in:

tooltip_keys

dionhaefner:
One more thing: Could you update the README with a new screenshot when you're done? :)

terracotta/client/app/src/sidebar-datasets/DatasetPreview.tsx
```diff
@@ -65,8 +64,8 @@ const DatasetPreview: FC<Props> = ({
            }, []).join('')
 
     return (
-        <TableRow>
-            <TableCell style={{ padding: 0 }} colSpan={8}>
+        <TableRow style={{height: 0}}>
+            <TableCell style={{ padding: 0, height: 'unset' }} colSpan={8}>
```
rihr, line 68: Could have put these in classes?

rihr (approved):

dionhaefner (commented):
Thanks! This works so much better!

### One last bug

Folding away the side bar is buggy - sometimes it works, sometimes this happens:

https://user-images.githubusercontent.com/11994217/116433799-2236e000-a84a-11eb-9067-0d8d00cab19f.mov

### Things I would do differently

- The new street view icon on the left is the wrong way around in my opinion. If you are on street view (default), the icon should be a satellite (to show that you can switch to satellite view by clicking on it). If you're on satellite, it should be a map.
- I would set something like `width: 100%` on the Terracotta logo to prevent it from being cut off.
- I personally really miss being able to enter the slider ranges. Maybe we can get a third opinion here? @j08lue ?
- By default, all sliders are set to full size. This leads to very low contrast in most cases. A IMO saner default would be to set it to the 5th and 95th percentile (you get this from `/metadata`).
- I would move the `view full metadata` link to after the JSON dict with the truncated metadata, maybe like this:

   Screen Shot 2021-04-28 at 18 04 41

- Resizing the sidebar to something very small looks weird, so ideally it would have a minimum size below which it snaps to 0 (this would also solve the bug above).

(these are just my opinion which you can ignore if you need to)

bertearazvan:
- [ ] toggle sidebar if it's too small
- [ ] input ranges for the sliders
- [ ] set sliders to valid procentage
- [ ] move "view metadata"

bertearazvan:
> * I would set something like `max-width: 100%` on the Terracotta logo to prevent it from being cut off.

If I would set it to `100%` then it would be huge if it's resized a lot. I believe that the current size is correct and I like the alignment to the left.

> * The old viewer would only zoom to the dataset if it wasn't already in view. For example: You look at Copenhagen on an image, then switch to a different image. You don't want it to zoom back to whole Denmark again.

The way it zooms to it now is based on the bounding box. So the map viewport will always wrap the current dataset selected with a slight padding from the margins of the map. I would not implement this functionality in this PR, but maybe an Issue could be rasied

dionhaefner (commented):
Can be removed?

terracotta/client/app/src/sidebar/VerticalHandle/VerticalHandle.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+	}, [  initialWidth ])
+
+	const handleMouseUp = useCallback(() => {
+
+		if (boxWidth < minWidth) {
+
+			onDrag(initialWidth)
+
+		}
```
dionhaefner, line 45: This can be removed?

dionhaefner (approved):
I could pick some more nits but I think it's nice now.

For the record: RGB slider ranges are not quite right (should take the percentiles of each band), and entering slider ranges is a bit flaky and has too many decimals IMO. But unless these are trivial fixes let's get this merged.

j08lue:
> For the record: RGB slider ranges are not quite right (should take the percentiles of each band), and entering slider ranges is a bit flaky and has too many decimals IMO. But unless these are trivial fixes let's get this merged.

Let us just add those as new issues for the future.

terracotta/client/app/src/sidebar/VerticalHandle/VerticalHandle.tsx
bertearazvan: For our case it's indeed not needed since it's `0`, but for reusability reasons, I would keep it. I'll instead pass it as optional props to that component.

PR #207, opened 2021-04-27
Test with python 3.9 / Ubuntu focal

dionhaefner:


dionhaefner:
I guess this is blocked until we can unpin rasterio.

dionhaefner:
Done in #242

PR #210, opened 2021-05-04
Test some tweaks to react app

dionhaefner:


dionhaefner:
What do you think of this @bertearazvan ? 😄

terracotta/client/app/src/sidebar-datasets/SidebarDatasets.tsx
```diff
@@ -155,7 +155,7 @@ const SidebarDatasetsItem: FC<Props> = ({
             if(dataset){
 
                 const { percentiles } = dataset
-                const validRange = [percentiles[5], percentiles[95]]
+                const validRange = [percentiles[4], percentiles[94]]
```
bertearazvan, line 158: I was actually wondering about this. Initially, I thought the same, that the 5th and 95th percentiles were actually the 4th and the 94th because the actual 0 and 100 were in the `range` key. Is that the reason?

If it is, I am wondering why they are not also added in the percentiles also... it would be much easier to spot that the same values are in the beginning and in the end of that array.

bertearazvan (approved):
I think it looks and behaves very well this way. Great job! 🥇

terracotta/client/app/src/sidebar-datasets/SidebarDatasets.tsx
dionhaefner: Yes, that's right. Maybe it would have been better design to add the min and the max as you suggest, but it's too late to fix this now. Breaking changes to the response API should only happen for a very good reason.

dionhaefner:
Call me mr. full stack 😎

PR #215, opened 2021-06-13
Update settings.rst

chapmanjacobd:
"['*']" doesn't work but this does '["*"]' (in fish shell at least, not sure about other shells)

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -73,7 +73,7 @@ endpoints individually with the settings :attr:`~terracotta.config.TerracottaSet
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


j08lue:
Yes, since this is supposed to be JSON, your corrected version must be the one. Thanks!

j08lue:
CI is failing because of MyPy not finding types for some of our dependencies. No idea why it started doing that.

https://mypy.readthedocs.io/en/stable/running_mypy.html#library-stubs-not-installed

Seems to be easily fixed in the build step with
```
mypy --install-types
```

dionhaefner:
Thanks @chapmanjacobd

Issue #171, opened 2020-03-30
Use click-spinner to visualize optimize-rasters progress


dionhaefner:
Finally! I've been looking for something like this for a while.

https://github.com/click-contrib/click-spinner

PR #228, opened 2021-08-06
Parallel raster optimisation

### LINKED ISSUE #55: Add parallel preprocessing capabilities
Preprocessing is pretty slow on large rasters. Processing several blocks in parallel could mitigate that. Alternatively, we can process multiple files in parallel (e.g. in `optimize_rasters`).

nickeopti:
Use a `multiprocessing.Pool` to optimize raster files in parallel (each raster file is applied individually). Closes #55 

Currently uses `multiprocessing.cpu_count()` number of threads. This seems to be the number of logical (hyper-threading) cores: I'm unsure whether the number of physical cores might be preferable.

Currently has some caveats:
1: The sub-progress bars are removed, such that only the overall progress is shown.
(And the filename postfix is now showing the just-completed file, rather than the currently-being-processed file).

2: When raising an exception because optimized files already exists (and neither  `--overwrite` nor `--skip-existing` flags are set), the entire trace-stack is now shown, where it previously just showed a nice little message.

3: The `test_reoptimize` test in `tests/scripts/test_optimize_rasters.py` fails. Not because the code performs wrongly, but because of the way exceptions inside threads are handled.
When submitting a task to the `Pool`, an `error_callback` function is given, which receives any errors which occurs in the thread (and handles them in the main-thread). But apparently pytest recognizes that an error has been thrown inside the threads, and fails the test based on that.
If a preprocessed file already exists, and `terracotta optimize-rasters` is run without `--overwrite` or `--skip-existing` flags, then the expected behaviour is to throw an exception. But somehow pytest expects the exception to be raised differently, or something like that? 
And I can't figure out how to make the test accept the new exception flow, such that the test just tests whether the code handles preexisting optimized files properly, and not where/how the exception is raised.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,6 +60,7 @@ setup(
         'apispec-webframeworks',
         'cachetools>=3.1.0',
         'click',
+        'click-spinner',
         'flask',
         'flask_cors',
         'marshmallow>=3.0.0',
--- a/terracotta/scripts/optimize_rasters.py
+++ b/terracotta/scripts/optimize_rasters.py
@@ -117,15 +117,10 @@ def _optimize_single_raster(
 ) -> None:
     output_file = output_folder / input_file.with_suffix('.tif').name
 
-    if output_file.is_file():
-        if skip_existing:
-            if not quiet:
-                click.echo(f'\r{input_file.name!r} {progress_suffix}')
-            return
-        if not overwrite:
-            raise click.BadParameter(
-                f'Output file {output_file!s} exists (use --overwrite or --skip-existing)'
-            )
+    if output_file.is_file() and skip_existing:
+        if not quiet:
+            click.echo(f'\r{input_file.name!r} skipped {progress_suffix}')
+        return
 
     if not quiet:
         click.echo(f'\r{input_file.name} ... {progress_suffix}')
@@ -265,6 +260,12 @@ def optimize_rasters(raster_files: Sequence[Sequence[Path]],
     for f in raster_files_flat:
         if not f.is_file():
             raise click.BadParameter(f'Input raster {f!s} is not a file')
+
+        output_file = output_folder / f.with_suffix('.tif').name
+        if output_file.is_file() and not (skip_existing or overwrite):
+            raise click.BadParameter(
+                f'Output file {f!s} exists (use --overwrite or --skip-existing)'
+            )
 
         with rasterio.open(str(f), 'r') as src:
             if src.count > 1 and not quiet:
--- a/terracotta/scripts/optimize_rasters.py
+++ b/terracotta/scripts/optimize_rasters.py
@@ -288,13 +288,33 @@ def optimize_rasters(raster_files: Sequence[Sequence[Path]],
 
         if cores == -1:
             cores = os.cpu_count() or 1  # Default to 1 if `cpu_count` returns None
-        with concurrent.futures.ProcessPoolExecutor(max_workers=cores) as executor:
-            pbar.write(f'Optimizing {len(raster_files_flat)} files on {cores} '
-                       f'process{"es" if cores > 1 else ""}...')
+        if cores > 1:
+            with concurrent.futures.ProcessPoolExecutor(max_workers=cores) as executor:
+                pbar.write(f'Optimizing {len(raster_files_flat)} files on {cores} processes...')
 
-            futures = [
-                executor.submit(
-                    _optimize_single_raster,
+                futures = [
+                    executor.submit(
+                        _optimize_single_raster,
+                        input_file,
+                        output_folder,
+                        skip_existing,
+                        overwrite,
+                        reproject,
+                        rs_method,
+                        in_memory,
+                        compression
+                    )
+                    for input_file in raster_files_flat
+                ]
+
+                for future in concurrent.futures.as_completed(futures):
+                    file_name, pixel_count, action = future.result()
+                    if not quiet:
+                        pbar.write(f"{action.capitalize()} {file_name!r}")
+                    pbar.update(pixel_count)
+        else:  # Single-core; run in the current process
+            for input_file in raster_files_flat:
+                file_name, pixel_count, action = _optimize_single_raster(
                     input_file,
                     output_folder,
                     skip_existing,
@@ -302,13 +322,7 @@ def optimize_rasters(raster_files: Sequence[Sequence[Path]],
                     reproject,
                     rs_method,
                     in_memory,
-                    compression,
+                    compression
                 )
-                for input_file in raster_files_flat
-            ]
-
-            for future in concurrent.futures.as_completed(futures):
-                file_name, pixel_count, action = future.result()
-                if not quiet:
-                    pbar.write(f"{action.capitalize()} {file_name!r}")
+                pbar.set_postfix(file=f"{file_name!r} ({action})")
                 pbar.update(pixel_count)
--- a/terracotta/scripts/optimize_rasters.py
+++ b/terracotta/scripts/optimize_rasters.py
@@ -289,21 +289,20 @@ def optimize_rasters(raster_files: Sequence[Sequence[Path]],
 
     output_folder.mkdir(exist_ok=True)
 
+    if nproc == -1:
+        nproc = os.cpu_count() or 1  # Default to 1 if `cpu_count` returns None
+
     if not quiet:
-        # insert newline for nicer progress bar style
-        click.echo('')
+        files_str = 'file' if len(raster_files_to_optimize) == 1 else 'files'
+        processes_str = 'process' if nproc == 1 else 'processes'
+        click.echo(f'Optimizing {len(raster_files_to_optimize)} {files_str} on {nproc} {processes_str}')
 
     with contextlib.ExitStack() as outer_env:
         outer_env.enter_context(click_spinner.spinner(beep=False, disable=False, force=False, stream=sys.stdout))
         outer_env.enter_context(rasterio.Env(**GDAL_CONFIG))
 
-        if nproc == -1:
-            nproc = os.cpu_count() or 1  # Default to 1 if `cpu_count` returns None
         if nproc > 1:
             executor = outer_env.enter_context(concurrent.futures.ProcessPoolExecutor(max_workers=nproc))
-            if not quiet:
-                click.echo(f'\rOptimizing {len(raster_files_to_optimize)} files '
-                            f'on {nproc} processes')
 
             futures = [
                 executor.submit(
--- a/terracotta/scripts/optimize_rasters.py
+++ b/terracotta/scripts/optimize_rasters.py
@@ -245,7 +245,7 @@ def optimize_rasters(raster_files: Sequence[Sequence[Path]],
 
     Note that all rasters may only contain a single band.
     """
-    raster_files_flat = sorted(set(itertools.chain.from_iterable(raster_files)))
+    raster_files_flat = set(itertools.chain.from_iterable(raster_files))
 
     if not raster_files_flat:
         click.echo('No files given')
@@ -257,15 +257,19 @@ def optimize_rasters(raster_files: Sequence[Sequence[Path]],
         compression = _prefered_compression_method()
 
     total_pixels = 0
+    raster_files_to_skip = set()
     for f in raster_files_flat:
         if not f.is_file():
             raise click.BadParameter(f'Input raster {f!s} is not a file')
 
         output_file = output_folder / f.with_suffix('.tif').name
-        if output_file.is_file() and not (skip_existing or overwrite):
-            raise click.BadParameter(
-                f'Output file {f!s} exists (use --overwrite or --skip-existing)'
-            )
+        if output_file.is_file():
+            if not (skip_existing or overwrite):
+                raise click.BadParameter(
+                    f'Output file {f!s} exists (use --overwrite or --skip-existing)'
+                )
+            elif skip_existing:
+                raster_files_to_skip.add(f)
 
         with rasterio.open(str(f), 'r') as src:
             if src.count > 1 and not quiet:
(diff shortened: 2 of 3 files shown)


j08lue:
We often use [`concurrent.futures.ProcessPoolExecutor`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor) these days. Often with `executor.map(worker, iterable)`. It has a slightly simpler interface than `multiprocessing`, and maybe a bit better error propagation?

In any case, perhaps worth a try to change to `concurrent.futures` and see how it goes with the exception from there.

dionhaefner:
Using `concurrent.futures` fixes problems 2 and 3. Exception propagation is notoriously hard in concurrent contexts, and `concurrent.futures` handles this nicely.

For problem 1, I suggest to replace the progress bars with a spinner (#171) and log messages (and a `--quiet` option or so). So something like

```bash
$ terracotta optimize-rasters *.tif
Optimizing 17 files on 5 processes
foo1.tif ...  (1/17)
foo2.tif ...  (2/17)
foo3.tif ...  (3/17)
foo4.tif ...  (4/17)
foo5.tif ...  (5/17)
<spinner>
```


But I think the real problem here is memory usage. We either need some heuristic that ensures we don't over-commit, or disable concurrency by default. Otherwise we risk crashing people's machines.

BTW, since all of the heavy lifting happens inside GDAL, I suggest you also try using a thread pool instead of processes. Chances are, it's the same speed with less overhead.

nickeopti:
The branch is now updated to use a `ProcessPoolExecutor` instead.

On my test data, the optimization process takes somewhere between 1:00 and 1:10 (min:sec) when using a  `ProcessPoolExecutor`, whereas using a  `ThreadPoolExecutor` increases the processing time to somewhere between 1:30 and 2:30 (varies a lot, sometimes at almost 3 minutes). But if you prefer a  `ThreadPoolExecutor` anyways, the change required is only that class exchange (in line 292).

The process indicator now shows
```
Optimizing 10 files on 8 processes...
Optimized 'file_1.tif'
Optimized 'file_2.tif'
Optimized 'file_3.tif'
Optimized 'file_4.tif'
Optimizing rasters:  40%|███████████████████████████████████████████████████████
```
where the progress bar shows the overall progress.

I still cannot figure out how to find out which files are currently being processed (i.e., how to be notified when the pool starts processing a new file) -- only which files are done being processed. Hence the 'Optimized' when the file is completed, rather than an 'Optimizing...' as you showed in your example.

A `--cores` flag is added, to manually set the number of cores to use. It now defaults to 1. Setting it to `-1` uses the (logical) CPU core count. Let me know if you a more elegant solution for this.
Also, if you have any suggestions regarding memory usage heuristics, that would be interesting. Otherwise, I guess this current solution works.

I have added the `--cores` flag to the `test_optimize_rasters` test. I hope that is in accordance with the style.
Even though it seems somewhat pointless to test multi-process optimization, when only a single file is supplied to the test method, I guess the test works anyways.

dionhaefner:
Let's stay with processes then.

I like it better when file being optimzed is printed. This makes it a lot easier to debug hangs or running out of memory (because you know which file is the culprit).

For this I would simply move the print to the subprocesses.

Also, could we switch the progress bar for a spinner? When optimizing large rasters this can hang for a long time, so I would like to give users the feeling that *something* is happening.

Here's a snipped to get you started:

```python
import time
import click
import click_spinner


@click.command()
def main():
    with click_spinner.spinner():
        i = 0
        while True:
            click.echo(f"\rfoo {i}")
            time.sleep(4)
            i += 1


if __name__ == "__main__":
    main()
```

nickeopti:
It now prints something like
```
Optimizing 10 files on 4 processes...
Optimizing 'file1.tif'... (1/10)
Optimizing 'file2.tif'... (2/10)
Optimizing 'file3.tif'... (3/10)
Optimizing 'file4.tif'... (4/10)
Optimized 'file4.tif'
Optimizing 'file5.tif'... (5/10)
Optimized 'file3.tif'
Optimizing 'file6.tif'... (6/10)
Optimized 'file2.tif'
Optimizing 'file7'... (7/10)
<spinner>
```
while optimising the raster files.

Does that match your intended style?

dionhaefner:
It's sort of a nitpick, but could we match the style of https://github.com/DHI-GRAS/terracotta/pull/228#issuecomment-894493670 exactly?

j08lue:
> exactly

Do you want the dots and `(n/m)` horizontally aligned, even when file names have different length? And from what file name length should names be truncated - if at all? 😉

dionhaefner:
No alignment or truncation or fancy stuff. Just don't repeat "optimizing" and don't log when you're done :)

terracotta/scripts/optimize_rasters.py
```diff
@@ -141,6 +220,12 @@ def _named_tempfile(basedir: Union[str, Path]) -> Iterator[str]:
     '--compression', default='auto', type=click.Choice(['auto', 'deflate', 'lzw', 'zstd', 'none']),
     help='Compression algorithm to use [default: auto (ZSTD if available, DEFLATE otherwise)]'
 )
+@click.option(
+    '--cores', default=1, type=click.INT,
+    help='Number of processes to use for multi-core processing '
+         '[default: 1, i.e., single-core processing] '
+         'Set to -1 to use all available (logical) cores'
+)
```
dionhaefner, line 228: I would prefer `--nproc` for this

tests/scripts/test_optimize_rasters.py
```diff
@@ -53,6 +55,9 @@ def test_optimize_rasters(unoptimized_raster_file, tmpdir, in_memory, reproject,
     if reproject:
         flags.append('--reproject')
 
+    if cores is not None:
+        flags.append('--cores=%i' % cores)
```
dionhaefner, line 59: Use an f-string

terracotta/scripts/optimize_rasters.py
```diff
@@ -194,94 +280,48 @@ def optimize_rasters(raster_files: Sequence[Sequence[Path]],
...
-    )
-
-    with contextlib.ExitStack() as outer_env:
-        pbar = outer_env.enter_context(tqdm.tqdm(
-            total=total_pixels, smoothing=0, disable=quiet,
-            bar_format='{l_bar}{bar}| [{elapsed}<{remaining}{postfix}]',
-            desc='Optimizing rasters'
-        ))
+    with contextlib.ExitStack() as outer_env, \
+         click_spinner.spinner(beep=False, disable=False, force=False, stream=sys.stdout):
```
dionhaefner, line 284: Just use the ExitStack to enter the spinner

terracotta/scripts/optimize_rasters.py
```diff
@@ -194,94 +280,48 @@ def optimize_rasters(raster_files: Sequence[Sequence[Path]],
...
-                    continue
-                if not overwrite:
-                    raise click.BadParameter(
-                        f'Output file {output_file!s} exists (use --overwrite or --skip-existing)'
+        if cores == -1:
+            cores = os.cpu_count() or 1  # Default to 1 if `cpu_count` returns None
+        if cores > 1:
+            with concurrent.futures.ProcessPoolExecutor(max_workers=cores) as executor:
+                if not quiet:
+                    click.echo(f'\rOptimizing {len(raster_files_flat)} files '
```
dionhaefner, line 292: Move this to before entering the ExitStack (then you don't need `\r` either)

terracotta/scripts/optimize_rasters.py
dionhaefner, line 284: `disable=quiet`

terracotta/scripts/optimize_rasters.py
```diff
@@ -194,94 +280,48 @@ def optimize_rasters(raster_files: Sequence[Sequence[Path]],
...
-            if output_file.is_file():
-                if skip_existing:
-                    continue
-                if not overwrite:
-                    raise click.BadParameter(
-                        f'Output file {output_file!s} exists (use --overwrite or --skip-existing)'
+        if cores == -1:
+            cores = os.cpu_count() or 1  # Default to 1 if `cpu_count` returns None
+        if cores > 1:
+            with concurrent.futures.ProcessPoolExecutor(max_workers=cores) as executor:
```
dionhaefner, line 290: Use ExitStack to save one level of indentation

nickeopti:
Updated to:
```
Optimizing 10 files on 4 processes
file1.tif ... (1/10)
file2.tif ... (2/10)
file3.tif ... (3/10)
file4.tif ... (4/10)
file5.tif ... (5/10)
<spinner>
```


> I like it better when the file currently being optimzed is printed. This makes it a lot easier to debug hangs or running out of memory (because you know which file is the culprit).

In order to know which files might be causing problems, we need to know which files are currently being processed, right? As the optimization of files probably won't complete in order, I thought we needed either
1. to log when a file is done, or
2. to remove the files from the list once they are completed.

I considered the second option -- I think it would be a nice solution -- but deemed it overkill.
A third option is of course to do neither; and just don't really know which files are currently being processed.

Btw, occasionally the logging of files happens slightly out-of-order (file2 might log before file1), when both processes are started almost-simultaneously. Then the order seems slightly odd (prints `(2/10)` just before `(1/10)`). Do we agree that that is unimportant, and just ignore it? Otherwise I think a synchronised counter is needed, and that seems a bit excessive.

terracotta/scripts/optimize_rasters.py
```diff
@@ -194,94 +280,48 @@ def optimize_rasters(raster_files: Sequence[Sequence[Path]],
...
+                if not quiet:
+                    click.echo(f'\rOptimizing {len(raster_files_flat)} files '
+                               f'on {cores} processes')
```
dionhaefner, line 293: Why is this not covered? 🤔

terracotta/scripts/optimize_rasters.py
```diff
@@ -194,94 +280,48 @@ def optimize_rasters(raster_files: Sequence[Sequence[Path]],
...
-                        dst, str(output_file), copy_src_overviews=True,
-                        compress=compression, **COG_PROFILE
-                    )
-
-            pbar.update(dst.height * dst.width)
+                    for i, input_file in enumerate(raster_files_flat, start=1)
+                ]
+
+                for future in concurrent.futures.as_completed(futures):
+                    future.result()  # Needed to throw any exceptions
```
dionhaefner, line 313: How about we do something like

```python
try:
    future.result()
except Exception as exc:
    raise RuntimeError(f'Error while optimizing file {filename}') from exc
```

Then we know which file acted up. For this you need a mapping of futures to filenames (you can use a dict, see [e.g. here](https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor-example)).

dionhaefner (commented):
Looks really good already, let's just get this polished a little more :)

dionhaefner:
> Btw, occasionally the logging of files happens slightly out-of-order (file2 might log before file1), when both processes are started almost-simultaneously. Then the order seems slightly odd (prints (2/10) just before (1/10)). Do we agree that that is unimportant, and just ignore it? Otherwise I think a synchronised counter is needed, and that seems a bit excessive.

Yes, no problem.

setup.py
```diff
@@ -60,6 +60,7 @@
         'apispec-webframeworks',
         'cachetools>=3.1.0',
         'click',
+        'click-spinner',
```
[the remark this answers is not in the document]
nickeopti: That is done in the commit f757591.

One disadvantage, though, is that `click.progressbar` does not have a `disable` flag to set when in `--quiet` mode.
Another is that the progressbar often isn't shown at all, as it is designed to only show if each task takes at least a second. See https://click.palletsprojects.com/en/8.0.x/api/#click.progressbar

Is removing a dependency enough to outweigh those disadvantages?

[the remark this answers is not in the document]
dionhaefner: Yea maybe not. Let's stick with TQDM. Thanks for exploring this.

terracotta/scripts/optimize_rasters.py
```diff
@@ -108,6 +110,83 @@ def _named_tempfile(basedir: Union[str, Path]) -> Iterator[str]:
...
+
+    if output_file.is_file():
+        if skip_existing:
+            if not quiet:
+                click.echo(f'\r{input_file.name!r} {progress_suffix}')
+            return
+        if not overwrite:
+            raise click.BadParameter(
+                f'Output file {output_file!s} exists (use --overwrite or --skip-existing)'
+            )
```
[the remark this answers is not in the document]
nickeopti: Done in commit 8d66a91. Should we also skip files early (such that we do not spin a task into a process, just to skip it)?

BTW, if someone provided both `--skip-existing` and `-overwrite`, the current behaviour is to just skip (skipping takes priority). Is that as intended? Or should we fail with a message to only provide one flag?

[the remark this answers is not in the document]
dionhaefner: > Should we also skip files early (such that we do not spin a task into a process, just to skip it)?

Yes, good idea. So just say "processing 8 files" if 2 are skipped due to existing already.

> BTW, if someone provided both --skip-existing and -overwrite, the current behaviour is to just skip (skipping takes priority). Is that as intended? Or should we fail with a message to only provide one flag?

Also good idea :)

nickeopti:
Status update:
I believe all change requests are now (finally) implemented.
Codecov is complaining that exception-handling (in the processes) isn't tested (lines 327-328). If you know how to trigger an exception there, please let me know. Otherwise I think we might have to just accept that it isn't tested.

dionhaefner:
To trigger an exception you can use [monkeypatch](https://docs.pytest.org/en/6.2.x/monkeypatch.html) to replace `_optimize_single_raster` with a function that raises an exception.

dionhaefner:
Thanks! This turned out to be quite a bit of work, but I think it was worth it. This has much better UX than before.

PR #229, opened 2021-08-06
Require pymysql>=1.0.0, which has a context managed connection

### LINKED ISSUE #185: pymysql 0.1.0 breaks context manager
In version 0.1.0, `pymysql.connect` is not a context manager, so our `with` won't work. 

https://github.com/DHI-GRAS/terracotta/blob/ec351a05fd4fb76133219d62fdd1f5450f7abfff/terracotta/drivers/mysql.py#L260

There is an open PR fixing that https://github.com/PyMySQL/PyMySQL/pull/886. Until it is merged, users should `pip install pymysql==0.9.3`.

nickeopti:
Closes #185, as pymysql version >=1.0 has a context managed connection. See https://github.com/PyMySQL/PyMySQL/blob/main/pymysql/connections.py#L354

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,12 +94,12 @@ setup(
             'sphinx',
             'sphinx_autodoc_typehints',
             'sphinx-click',
-            'pymysql<0.10'
+            'pymysql>=1.0.0'
         ],
         'recommended': [
             'colorlog',
             'crick',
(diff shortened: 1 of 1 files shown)


j08lue (approved):
Thanks!

PR #230, opened 2021-08-10
Introduce multiqueries for dataset lookup #33

### LINKED ISSUE #33: Introduce multiqueries for dataset lookup
To make it easier to scale to large data collections, we should support queries in `/datasets` such as

`/datasets?year=[2016,2018]`

which would return all datasets from `2016` and `2018`.

Another consideration could be range-based queries, but that would require the introduction of per-key datatypes, which is something I'd like to avoid for now.

nickeopti:
Make it possible to query datasets like `/datasets?year=[2016,2018]&band=02`. This closes #33.

The parsing of lists of values is implemented rather naively, but it works.

The `driver.get_datasets()` method now accepts `where: Mapping[str, Union[str, List[str]]]`. Handling both cases (conditions either string or list of strings) have introduced some somewhat convoluted code to create the SQL queries, but I think it's about as good as it gets without a larger rewrite.

### DIFF
--- a/terracotta/drivers/mysql.py
+++ b/terracotta/drivers/mysql.py
@@ -4,11 +4,12 @@ MySQL-backed raster driver. Metadata is stored in a MySQL database, raster data 
 to be present on disk.
 """
 
-from typing import (Tuple, Dict, Iterator, Sequence, Union,
+from typing import (List, Tuple, Dict, Iterator, Sequence, Union,
                     Mapping, Any, Optional, cast, TypeVar)
 from collections import OrderedDict
 import contextlib
 from contextlib import AbstractContextManager
+import itertools
 import re
 import json
 import urllib.parse as urlparse
@@ -335,7 +336,7 @@ class MySQLDriver(RasterDriver):
     @trace('get_datasets')
     @requires_connection
     @convert_exceptions('Could not retrieve datasets')
-    def get_datasets(self, where: Mapping[str, str] = None,
+    def get_datasets(self, where: Mapping[str, List[str]] = None,
                      page: int = 0, limit: int = None) -> Dict[Tuple[str, ...], str]:
         cursor = self._cursor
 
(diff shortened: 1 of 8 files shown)


terracotta/drivers/mysql.py
```diff
@@ -354,10 +355,19 @@ def get_datasets(self, where: Mapping[str, str] = None,
             if not all(key in self.key_names for key in where.keys()):
                 raise exceptions.InvalidKeyError('Encountered unrecognized keys in '
                                                  'where clause')
-            where_fragment = ' AND '.join([f'{key}=%s' for key in where.keys()])
+            where = {
+                key: value if isinstance(value, list) else [value]
```
dionhaefner, line 359: I would prefer a

```python
if isinstance(value, str):
    value = [value]
```

terracotta/drivers/mysql.py
```diff
@@ -354,10 +355,19 @@ def get_datasets(self, where: Mapping[str, str] = None,
...
+                for key, value in where.items()
+            }
+            conditions = [
+                '(%s)' % ' OR '.join([f'{key}=%s'] * len(value))
+                for key, value in where.items()
+            ]
+            values = list(itertools.chain(*where.values()))
+            where_fragment = ' AND '.join(conditions)
```
dionhaefner, line 367: This is going a bit too far with the comprehensions. I think a straightforward loop might be more readable?

terracotta/drivers/sqlite.py
```diff
@@ -252,10 +253,19 @@ def get_datasets(self, where: Mapping[str, str] = None,
...
+            where = {
+                key: value if isinstance(value, list) else [value]
+                for key, value in where.items()
+            }
+            conditions = [
+                '(%s)' % ' OR '.join([f'{key}=?'] * len(value))
+                for key, value in where.items()
+            ]
+            values = list(itertools.chain(*where.values()))
+            where_fragment = ' AND '.join(conditions)
```
dionhaefner, line 265: Same here

dionhaefner (commented):
Looks good, some minor changes, and please add a test for this.

terracotta/drivers/sqlite.py
j08lue: A delicate detail of style in code bases - the cases in which list comprehensions are used and where not. 😄I think we generally keep it down to very simple cases. Readability of the code is the property to optimize for.

For readability, I also prefer format strings over the old `%s` notation...

dionhaefner (approved):
Looks good to me, thanks!

PR #234, opened 2021-09-02
Issue104; get valid key value combinations

### LINKED ISSUE #104: Supply a way to retrieve unique key values
When serving heterogeneous datasets (e.g. data with a `type` key), if could be very useful to be able to retrieve all unique key values.

Examples:

1. Get all unique keys
  ```
  >>> /unique_endpoint
  {
      'type': ['reflectance', 'index', 'composite'],
      'date': ['20180101', '20180101_20180201'],
      'id': ['1', '2', '3']
  }
  ```

2. Get keys valid for reflectances
  ```
  >>> /unique_endpoint?type=reflectance
  {
      'type': ['reflectance'],
      'date': ['20180101'],
      'id': ['1']
  }
  ```

3. Get all data for a specific date
  ```
  >>> /unique_endpoint?date=20180101
  {
      'type': ['reflectance', 'index'],
      'date': ['20180101'],
      'id': ['1', '2']
  }
  ```

An alternative could be to find a better abstraction / separation between datasets.

nickeopti:
Make it possible to discover which key values exists -- optionally after filtering for specific key values.
This closes #104.

Can be called as
```
>>> /valid_values
{
    'region': ['a', 'b, 'c'],
    'date': ['20180101', '20190101', '20200101'],
    'band': ['E', 'T', 'I']
}
```
to get all possible values for each key,
or
```
>>> /valid_values?region=a
{
    'region': ['a'],
    'date': ['20180101', '20190101', '20200101'],
    'band': ['1', '2', '3']
}
```
to get all valid values for each key, when one or more keys are set,
or
```
>>> /valid_values?region=a&band=[1,2]
{
    'type': ['reflectance'],
    'date': 'date': ['20180101', '20190101'],
    'band': ['1', '2']
}
```
where, in the last query, notice that the multi-querying syntax introduced from issue #33 is supported.

### DIFF
--- a/terracotta/drivers/base.py
+++ b/terracotta/drivers/base.py
@@ -82,6 +82,10 @@ class Driver(ABC):
         pass
 
     @abstractmethod
+    def get_valid_values(self, where: Mapping[str, Union[str, List[str]]]) -> Dict[str, List[str]]:
+        pass
+
+    @abstractmethod
     def get_datasets(self, where: Mapping[str, Union[str, List[str]]] = None,
                      page: int = 0, limit: int = None) -> Dict[Tuple[str, ...], Any]:
         # Get all known dataset key combinations matching the given constraints,
--- a/terracotta/drivers/mysql.py
+++ b/terracotta/drivers/mysql.py
@@ -332,6 +332,35 @@ class MySQLDriver(RasterDriver):
 
         return out
 
+    @requires_connection
+    @convert_exceptions('Could not retrieve valid key values')
+    def get_valid_values(self, where: Mapping[str, Union[str, List[str]]]) -> Dict[str, List[str]]:
+        cursor = self._cursor
+
+        if not all(key in self.key_names for key in where.keys()):
+            raise exceptions.InvalidKeyError('Encountered unrecognized keys in where clause')
+
+        conditions = []
+        values = []
+        for key, value in where.items():
+            if isinstance(value, str):
+                value = [value]
+            values.extend(value)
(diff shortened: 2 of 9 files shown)


dionhaefner:
Thanks for the implementation, looks pretty good. But I'm a bit hesitant to support this without a clear use case. Adding another SQL command complicates the drivers even more, so this looks a bit like feature creep to me.

What do you think @j08lue? Have you ever missed this feature?

j08lue:
> Have you ever missed this feature?

I am not sure. Not me, personally. Maybe @bertearazvan can comment whether we are doing this in any frontend:

1. Fetch all `/datasets` into the client
2. Scan datasets to get unique values for each key

If we do that anywhere, e.g. in order to create menus for selecting from them or so, this endpoint could replace that. But the result can be misleading - these are the unique values, but you do not know whether any combination of them exists until you check (e.g. `/datasets`).

bertearazvan:
We have the following use cases where this can be useful:

- When fetching all the dates available for an RGB raster. We always query for the `B01` band, which works. This I find would be a more "correct" approach.
- Usually very useful for fetching all the dates available for a product.
- In the ET4FAO app https://et4fao.dhigroup.com/#/, we are filtering through the available datasets that comply with the user's selection. This can be useful in this case too if there would be many more datasets, and it would be heavy to filter through them in the browser.

I think we can benefit from this in the future when dealing with a large number of datasets - @j08lue like LBST, maybe?

PR #242, opened 2021-12-15
Switch to GH actions

### LINKED ISSUE #204: Migrate CI from Travis to GitHub Actions?
Travis has served us well and it there are no indications that this would stop soon. But most new projects just use GitHub actions and many old migrate, due to the ease of having it all in one UI. Should we also make the switch?

Should be very easy to do...

What do you think, @dionhaefner?

dionhaefner:
Fixes #204


vincentsarago:
👋 @dionhaefner you might try using https://github.com/nektos/act to run action locally 😄

dionhaefner:
Indeed 😬

dionhaefner:
Except then I have to use the mess that is Docker on OSX ...

... almost ... there ...

vincentsarago:
@dionhaefner yes, you just have to make sure to use `--container-architecture linux/amd64` 🤷

dionhaefner:
whew

vincentsarago:
👏

Issue #245, opened 2022-01-04
Add database migration script


dionhaefner:
We've had the issue a couple of times where a new release breaks compatibility with the database format while not really changing much. The way this is handled in other libraries is by providing a migration script that works like this:

```bash
$ terracotta migrate mysql://example.com
Found Terracotta database v0.7.3 (this is v0.8.0)
Bump database to v0.8.0? (y/N)
Performing database migration ...
Done!

$ terracotta migrate mysql://example2.com
Found Terracotta database v0.0.1 (this is v0.8.0)
Error: cannot auto-migrate. Please re-create database.

$ terracotta migrate mysql://example3.com
Found Terracotta database v0.8.1 (this is v0.8.0)
Nothing to do
```

This requires some sort of incremental changelog that specifies the necessary transformations between versions.

j08lue:
That would be sweet. I had this case a few dozen times when upgrading deployments. I usually just bump the version in the DB manually when In know that the schema did not change. This would be a nice tool for that.

mrpgraae:
Since we are using SQLAlchemy, I would suggest using Alembic https://alembic.sqlalchemy.org/en/latest/

Alembic is a CLI tool, but they also have a python API, so you can call their migration commands from within your own tool.

dionhaefner:
Sounds good to me. Let's just make it an optional dependency that we check for inside the migration script.

dionhaefner:
Okay so I played around with alembic and ... didn't love it.

The biggest problem (apart from the steep learning curve) is that modifying SQLite is not fully supported (for example re-naming tables which just ends in `NotImplementedError`).

I'm hesitant whether we need a migration library at all. Some simple `ALTER TABLE` statements should do it. I suggest we use SQLalchemy to connect to the DB, then run the migration via `engine.execute` and raw SQL.

@nickeopti could you perhaps imagine writing the SQLalchemy boilerplate for this? :grimacing: Then I can finish up the actual migrations and CLI and such.

nickeopti:
I hope to get time to help at some point. Until then, this can probably get you started:

We can actually use the driver logic, just set `verify=False` to disable all checks:
```python
import sqlalchemy as sqla
from terracotta import drivers

driver = drivers.get_driver('<driver_path>')
with driver.connect(verify=False):
    driver.meta_store.connection.execute(sqla.text('<raw sql statement>'))
```

PR #248, opened 2022-01-31
Sqlalchemy compositional

nickeopti:
Implement the compositional `TerracottaDriver` structure proposed in https://github.com/DHI-GRAS/terracotta/pull/240#issuecomment-1013688976.

In a separate PR (going into #240) for better diffs.


nickeopti:
I really like this structure. The concept of lifting out functionality to the `TerracottaDriver` is really nice! This has made the code in `RelationalDriver` (in particular) and `RasterDriver` a bit leaner. 

However, I have encountered two problems, which I need help to fix. 
First is CI related; the tests fail in `Initialize mypy`.
And when we fix that, we will see another problem; something `pickle` related regarding `future`s in https://github.com/DHI-GRAS/terracotta/blob/sqlalchemy-compositional/terracotta/drivers/raster_base.py#L523 fails, which in turns make a lot of stuff fail. I hope this is an easy fix, but I have no clue where to start

dionhaefner:
Thanks for the implementation. I'll take care of the CI failures.

dionhaefner:
Is this ready for review @nickeopti ?

dionhaefner:
I'll review this soon, then. In the meantime, could you please bump the test coverage a bit? In general we don't accept PRs that lower global coverage.

nickeopti:
I've bumped test coverage a bit. There are now two uncovered branches left in this PR's diff -- neither of which I know how to hit.
The coverage also slightly decreases in #240, but I believe that is because the number of total lines of code decreases -- the PR's diff itself has 100% coverage.

dionhaefner (changes requested):
Thanks, this looks very nice!

Some minor issues but overall this looks like a huge improvement of the code structure.

terracotta/drivers/base.py
```diff
@@ -7,9 +7,11 @@
 import functools
```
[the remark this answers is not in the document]
nickeopti: What about `base_classes.py`?

terracotta/drivers/driver.py
```diff
@@ -0,0 +1,177 @@
...
+
+        metadata = self.metastore.get_metadata(keys)
+
+        if metadata is None:
+            # metadata is not computed yet, trigger lazy loading
+            handle = only_element(self.get_datasets(keys).values())
+            metadata = self.compute_metadata(handle, max_shape=self.LAZY_LOADING_MAX_SHAPE)
+            self.insert(keys, handle, metadata=metadata)
+
+            # this is necessary to make the lazy loading tests pass...
```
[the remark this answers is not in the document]
nickeopti: Probably because of small deviations in floating points... Alternatively because one of them (can't remember which) returned a list of tuples, whereas the other returned i list of lists. Not quite sure why, but that fixed a failing test.
But I'd certainly prefer to be without it -- it seems weird

terracotta/drivers/raster_base.py
```diff
@@ -3,33 +3,21 @@
 Base class for drivers operating on physical raster files.
```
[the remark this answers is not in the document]
nickeopti: or `geotiff_raster_store.py`?

terracotta/drivers/relational_base.py
```diff
@@ -10,17 +10,17 @@
 import urllib.parse as urlparse
```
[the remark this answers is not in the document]
nickeopti: or `relational_meta_store.py`?

terracotta/drivers/base.py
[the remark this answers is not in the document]
dionhaefner: yes :)

terracotta/drivers/relational_base.py
```diff
@@ -375,78 +349,55 @@ def get_metadata(self, keys: Union[Sequence[str], Mapping[str, str]]) -> Dict[st
...
 
         self.connection.execute(
             datasets_table
             .delete()
-            .where(*[datasets_table.c.get(column) == value for column, value in key_dict.items()])
+            .where(*[datasets_table.c.get(column) == value for column, value in keys.items()])
         )
         self.connection.execute(
-            datasets_table.insert().values(**key_dict, filepath=override_path)
+            datasets_table.insert().values(**keys, filepath=handle)
```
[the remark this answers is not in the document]
dionhaefner: I think keeping override_path is fine, because a handle always represents a path of some kind (be it a filepath or a path in a database).

terracotta/drivers/raster_base.py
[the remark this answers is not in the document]
dionhaefner: Whichever you prefer, just keep it consistent with the meta stores.

terracotta/drivers/relational_base.py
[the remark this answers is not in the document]
dionhaefner: Or should we even re-name `handle` to `path` globally? It's a bit more overloaded (there are several things that you might call "path" in Terracotta), but should be more intuitive for the most common use cases...

terracotta/drivers/driver.py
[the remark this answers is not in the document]
dionhaefner: I think it would pay off to find out which one it is. If it's float precision then we should absolutely keep this to keep things consistent. If it's different data structures we should just change the test to allow for that, since JSON responses don't distinguish between list and tuple.

terracotta/drivers/relational_base.py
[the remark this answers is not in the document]
j08lue: Thanks for the heads-up, @nickeopti. No problem to break the API here, so please go ahead naming things the way you two agree on.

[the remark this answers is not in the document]
nickeopti: I don't have strong opinions on this, other than I think we shall keep it consistent. If going for `handle`, then use `override_handle`, and so forth.

I like the abstraction of `handle`/`identifier`/`asset`/`reference` or similar, but if we never actually utilise that generalisation, then going for the more obvious and descriptive `path` also seems like a good choice.

terracotta/drivers/driver.py
```diff
@@ -0,0 +1,177 @@
...
+            raise exceptions.InvalidKeyError(
+                f'Encountered unrecognized keys {unknown_keys} (available keys: {self.key_names})'
+            )
+
+        # Apparently mypy thinks fun might still be None, hence the ignore:
+        return fun(self, keys, *args, **kwargs)  # type: ignore
+    return inner
+
+
+class TerracottaDriver:
```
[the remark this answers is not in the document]
nickeopti: Just to be sure:
Do you want to move all the docstrings away from `RasterStore` and `MetaStore`, or do you want to duplicate them in `TerracottaDriver`?

Also, related to this, I've wondered whether we should actually do
```python
class TerracottaDriver(MetaStore, RasterStore):
    ...
```
It somehow simultaneously seems so wrong and correct to me; it type checks, and makes use of `MetaStore` and `RasterStore` as interfaces. But perhaps it also goes a bit contra to the compositional structure?

terracotta/drivers/driver.py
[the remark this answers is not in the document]
dionhaefner: Let's just adjust the comment then :)

terracotta/drivers/relational_base.py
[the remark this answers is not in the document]
dionhaefner: Sounds reasonable. Let's do `path`, then.

terracotta/drivers/driver.py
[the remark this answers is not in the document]
dionhaefner: > Just to be sure:
> 
> Do you want to move all the docstrings away from `RasterStore` and `MetaStore`, or do you want to duplicate them in `TerracottaDriver`?

You want the "good" (public) docstrings on `TerracottaDriver`. The other classes can have "internal" (or no) docstrings.

> Also, related to this, I've wondered whether we should actually do
> 
> ```python
> 
> class TerracottaDriver(MetaStore, RasterStore):
> 
>     ...
> 
> ```
> 
> It somehow simultaneously seems so wrong and correct to me; it type checks, and makes use of `MetaStore` and `RasterStore` as interfaces. But perhaps it also goes a bit contra to the compositional structure?

Preposterous. Off to OOP jail with you 😤

mrpgraae:
@nickeopti @dionhaefner 
Is there any additional work needed before this can be merged? I want to get started on Postgres support, but it sort of depends on these changes.

dionhaefner:
We're still a bit light on documentation / proper docstrings in the right places. If you want to expedite the process you're welcome to take care of it :)

nickeopti:
I believe documentation is all that is remaining before merging this PR into #240. Then there's the question of whether to implement a (basic?) database migration tool #245 before merging #240 into `main`.

@mrpgraae have you begun improving the docstrings? I see you marked the change request as resolved. Otherwise I'll happily do it now. I've been busy with some other projects recently, but I have regained some time now :)

mrpgraae:
> Then there's the question of whether to implement a (basic?) database migration tool #245 before merging #240 into `main`.

That should probably be done in a different PR but before the next release.

>@mrpgraae have you begun improving the docstrings? I see you marked the change request as resolved. Otherwise I'll happily do it now. I've been busy with some other projects recently, but I have regained some time now :)

No sorry that must have been by mistake! I didn't do anything yet, but I'm happy to do it :slightly_smiling_face:

terracotta/drivers/geotiff_raster_store.py
```diff
@@ -0,0 +1,164 @@
...
+    def compute_metadata(self, path: str, *,
+                         extra_metadata: Any = None,
+                         use_chunks: bool = None,
+                         max_shape: Sequence[int] = None) -> Dict[str, Any]:
+        return raster.compute_metadata(path, extra_metadata=extra_metadata,
+                                       use_chunks=use_chunks, max_shape=max_shape,
+                                       large_raster_threshold=self._LARGE_RASTER_THRESHOLD,
+                                       rio_env=self._rio_env)
+
+    # return type has to be Any until mypy supports conditional return types
```
mrpgraae, line 99: I think this is just
```python
from typing import Union

def compute_metadata(...) -> Union[Future[np.ma.MaskedArray], np.ma.MaskedArray]
```

But in any case, having multiple return types sort of defeats the whole purpose of static types.
I can't see how a statically typed language could ever allow something like this without generics or metaprogramming.

The proper thing to do here would probably be to split this into 2 methods `compute_metadata` and `compute_metadata_async`. But I guess this might break the API elsewhere.

This is actual a general problem with async, that you basically have to implement your API twice if you want to have both an async and non-async API. There is a semi-famous rant about this problem here http://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/

mrpgraae: Anyways, if you are strict about mypy/static type hints, then using `Union` is usually a bit of a code smell.

Only exception is `Optional[SomeType]` of course, which is just shorthand for `Union[SomeType, None]`.

terracotta/drivers/terracotta_driver.py
```diff
@@ -0,0 +1,184 @@
...
+
+        return self.raster_store.get_raster_tile(
+            path=path,
+            tile_bounds=tile_bounds,
+            tile_size=tile_size,
+            preserve_values=preserve_values,
+            asynchronous=asynchronous,
+        )
+
+    def compute_metadata(self, path: str, *,
```
mrpgraae, line 128: Any reason why `compute_metadata` takes a path to identify the dataset, while `get_raster_tile` takes a sequence of keys? I would think they should both take keys, for consistency?

mrpgraae: Nevermind, I guess that `TerracottaDriver.get_metadata()` is the proper interface here.

But then I don't think there is any reason to expose `compute_metadata()` on `TerracottaDriver`?

mrpgraae:
Updated the docstrings now, but test collection is failing in CI now. I'm a bit puzzled as to what could cause this.

nickeopti:
> Updated the docstrings now, but test collection is failing in CI now. I'm a bit puzzled as to what could cause this.

The timing coincides with the release of pytest 7, which seems to have some breaking changes: https://docs.pytest.org/en/stable/changelog.html#breaking-changes

That may (partly) explain it. If that is the case, then the next question is whether we shall simply fix pytest to < 7 for now, or figure out what actually causes the errors. The changelog mentions something regarding potential breaking changes for plugins, so it may be caused by incompatibility with one of
```
pytest-cov
pytest-mypy
pytest-flake8
pytest-benchmark
```

mrpgraae:
> > Updated the docstrings now, but test collection is failing in CI now. I'm a bit puzzled as to what could cause this.
> 
> The timing coincides with the release of pytest 7, which seems to have some breaking changes: https://docs.pytest.org/en/stable/changelog.html#breaking-changes
> 
> That may (partly) explain it. If that is the case, then the next question is whether we shall simply fix pytest to < 7 for now, or figure out what actually causes the errors. The changelog mentions something regarding potential breaking changes for plugins, so it may be caused by incompatibility with one of
> 
> ```
> pytest-cov
> pytest-mypy
> pytest-flake8
> pytest-benchmark
> ```

Ah nice, good catch! Let's just pin to major version 6 for now. Then we can migrate to pytest 7 in a separate PR.

nickeopti:
Yep, the `pytest-flake8` and pytest 7 combination seems to be the culprit: https://github.com/tholo/pytest-flake8/issues/83. This problem applies to the entire project. Will you pin the version?

mrpgraae:
> Yep, the `pytest-flake8` and pytest 7 combination seems to be the culprit: [tholo/pytest-flake8#83](https://github.com/tholo/pytest-flake8/issues/83). This problem applies to the entire project. Will you pin the version?

:ok_hand: 
Pinned now.

terracotta/drivers/terracotta_driver.py
dionhaefner: It has to be exposed for this recipe:

```python
mtd = driver.compute_metadata(raster_file, extra_metadata={"custom": "foo"})
driver.insert(raster_file, metadata=mtd)
```

`compute_metadata` is used to compute *before* insertion, so keys are meaningless at this stage.

terracotta/drivers/geotiff_raster_store.py
dionhaefner: `Union` means "type could be either this or that", so you have to guard against both cases in downstream code. However, in this case, the return type is perfectly deterministic, based on the `async` input value.

> I can't see how a statically typed language could ever allow something like this without generics or metaprogramming.

```python
@typing.overload
def compute_metadata(..., async: TrueType) -> Future[np.ma.MaskedArray]:
    pass

@typing.overload
def compute_metadata(..., async: FalseType) -> np.ma.MaskedArray:
    pass
```

(pretty clean IMO, but not supported)

> The proper thing to do here would probably be to split this into 2 methods compute_metadata and compute_metadata_async. But I guess this might break the API elsewhere.

The code is clear and idiomatic as it is written. I see no reason to change that just because the type checker can't handle it. At this point, mypy has arguably done more harm than good to this project, by frequently breaking CI for no reason and exposing O(1) bugs at best. So I'm very hesitant to change the API just to make it happy.

dionhaefner:
The pytest-flake8 issue is fixed in #252 and will land in `main` soon.

terracotta/drivers/terracotta_driver.py
mrpgraae: Makes sense. I also did make sure to add documentation for this method.

terracotta/drivers/geotiff_raster_store.py
mrpgraae: I would say that "typing overload" is exactly what generics is, so this solution would be equivalent to using generics. Just different syntax :slightly_smiling_face: 

I don't think it's worth changing either. Just commenting on that comment in the code, that there is a way to have multiple return types with `Union`, but it's not really useful because it doesn't really help you, type-check-wise or code-completion-wise.

terracotta/drivers/terracotta_driver.py
```diff
@@ -0,0 +1,342 @@
...
+        self.LAZY_LOADING_MAX_SHAPE: Tuple[int, int] = settings.LAZY_LOADING_MAX_SHAPE
+
+    @property
+    def db_version(self) -> str:
+        """Terracotta version used to create the meta store.
+
+        Returns
+
+            A str specifying the version of Terracotta that was used to create the meta store.
+        """
```
dionhaefner, line 43:
```suggestion
        Returns:

            A str specifying the version of Terracotta that was used to create the meta store.

        """
```

terracotta/drivers/terracotta_driver.py
```diff
@@ -0,0 +1,342 @@
...
+T = TypeVar('T')
+
+
+def squeeze(iterable: Collection[T]) -> T:
+    assert len(iterable) == 1
+    return next(iter(iterable))
+
+
+class TerracottaDriver:
+
```
dionhaefner, line 28:
```suggestion
    """Terracotta driver object used to retrieve raster tiles and metadata.
    
    Do not instantiate directly, use :func:`terracotta.get_driver` instead.
    """
```

terracotta/drivers/terracotta_driver.py
```diff
@@ -0,0 +1,342 @@
...
+        return self.meta_store.db_version
+
+    @property
+    def key_names(self) -> Tuple[str, ...]:
+        """Get names of all keys defined by the meta store.
+
+        Returns:
+
+            A tuple defining the key names and order.
+        """
```
dionhaefner, line 53:
```suggestion

        """
```

terracotta/drivers/terracotta_driver.py
```diff
@@ -0,0 +1,342 @@
...
+        return self.meta_store.key_names
+
+    def create(self, keys: Sequence[str], *,
+               key_descriptions: Mapping[str, str] = None) -> None:
+        """Create a new, empty metastore.
```
dionhaefner, line 58:
```suggestion
        """Create a new, empty metadata store.
```

terracotta/drivers/terracotta_driver.py
```diff
@@ -0,0 +1,342 @@
...
+
+        Arguments:
+
+            keys: A sequence defining the key names and order.
+            key_descriptions: A mapping from key name to a human-readable
+                description of what the key encodes.
+        """
```
dionhaefner, line 65:
```suggestion

        """
```

terracotta/drivers/terracotta_driver.py
```diff
@@ -0,0 +1,342 @@
...
+        """Register a new dataset. Used to populate meta store.
+
+        Arguments:
+
+            keys: Keys of the dataset. Can either be given as a sequence of key values, or
+                as a mapping ``{key_name: key_value}``.
+            path: Path to access dataset (driver dependent).
+            override_path: If given, this path will be inserted into the meta store
+                instead of the one used to load the dataset.
+            metadata: Metadata for the dataset. If not given, metadata will be computed.
```
dionhaefner, line 192:
```suggestion
            metadata: Metadata dict for the dataset. If not given, metadata will be computed 
                via :meth:`compute_metadata`.
```

terracotta/drivers/terracotta_driver.py
```diff
@@ -0,0 +1,342 @@
...
+            skip_metadata: If True, will skip metadata computation, even if metadata is not given.
```
dionhaefner, line 193:
```suggestion
            skip_metadata: If True, will skip metadata computation (will be computed 
                during first request instead). Has no effect if ``metadata`` argument is given.
```

terracotta/drivers/terracotta_driver.py
```diff
@@ -0,0 +1,342 @@
...
+
+            - ``range``: global minimum and maximum value in dataset
+            - ``bounds``: physical bounds covered by dataset in latitude-longitude projection
+            - ``convex_hull``: GeoJSON shape specifying total data coverage in latitude-longitude
+              projection
+            - ``percentiles``: array of pre-computed percentiles from 1% through 99%
+            - ``mean``: global mean
+            - ``stdev``: global standard deviation
+            - ``metadata``: any additional client-relevant metadata
+        """
```
dionhaefner, line 289:
```suggestion

        """
```

terracotta/drivers/terracotta_driver.py
```diff
@@ -0,0 +1,342 @@
...
+                f'Encountered unrecognized keys {unknown_keys} (available keys: {self.key_names})'
+            )
+
+        return keys
+
+    def __repr__(self) -> str:
+        return (
+            f'{self.__class__.__name__}(\n'
+            f'\tmeta_store={self.meta_store.__class__.__name__}(path="{self.meta_store.path}"),\n'
+            f'\traster_store={self.raster_store.__class__.__name__}()\n'
```
dionhaefner, line 340: This should just call `repr(self.meta_store)` and `repr(self.raster_store)`.

dionhaefner (commented):
Some last nitpicks.

terracotta/drivers/sqlite_meta_store.py
```diff
@@ -8,10 +8,10 @@
...
 
-from terracotta.drivers.relational_base import RelationalDriver
+from terracotta.drivers.relational_meta_store import RelationalMetaStore
 
 
-class SQLiteDriver(RelationalDriver):
+class SQLiteDriver(RelationalMetaStore):
     """An SQLite-backed raster driver.
 
     Assumes raster data to be present in separate GDAL-readable files on disk or remotely.
```
dionhaefner, line 17: Nope

terracotta/drivers/sqlite_meta_store.py
```diff
@@ -8,10 +8,10 @@
 from pathlib import Path
 from typing import Union
 
-from terracotta.drivers.relational_base import RelationalDriver
+from terracotta.drivers.relational_meta_store import RelationalMetaStore
 
 
-class SQLiteDriver(RelationalDriver):
+class SQLiteDriver(RelationalMetaStore):
```
dionhaefner, line 14: IMO these classes should not be called `Driver` anymore. -> `SQLiteMetaStore`

terracotta/drivers/base_classes.py
```diff
@@ -136,19 +139,53 @@ def get_metadata(self, keys: Union[Sequence[str], Mapping[str, str]]) -> Dict[st
...
+               path: Any, **kwargs: Any) -> None:
+        """Register a new dataset. Used to populate metadata database.
+
+        Arguments:
+
+            keys: Keys of the dataset. Can either be given as a sequence of key values, or
+                as a mapping ``{key_name: key_value}``.
+            path: Path to access dataset (driver dependent).
+
+        """
```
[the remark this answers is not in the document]
mrpgraae: Yeah makes sense. I focused on just adding elaborate docs 'TerracottaDriver` as that is the outermost interface. I'll make sure that internal methods have short, descriptive docstrings.

dionhaefner:
More issues:

- [ ] Docs are out of date (but I can take care of them after everything else is stable)
- [ ] I would like to revisit whether `RemoteSQLite` should inherit from `SQLite`. I would argue that it should not inherit, because it increases complexity and saves you a handful lines of code at best, but if both of you prefer inheritance I'm OK with it.

mrpgraae:
> More issues:
> 
>     * [ ]  Docs are out of date (but I can take care of them after everything else is stable)
> 
>     * [ ]  I would like to revisit whether `RemoteSQLite` should inherit from `SQLite`. I would argue that it should not inherit, because it increases complexity and saves you a handful lines of code at best, but if both of you prefer inheritance I'm OK with it.

>Docs are out of date (but I can take care of them after everything else is stable)

I guess as long as they get updated before the next release, it should be fine. Maybe we can build them as part of the CI chain?

>I would like to revisit whether `RemoteSQLite` should inherit from `SQLite`. I would argue that it should not inherit, because it increases complexity and saves you a handful lines of code at best, but if both of you prefer inheritance I'm OK with it.

This must be a very subjective thing, because in my mind it's less complex in the current structure :smile: This way, I can look at `RemoteSQLite` and see it as a "diff" of how it behaves differently from the regular `SQLite` backend. If we did not have inheritance, there would be duplicate code and I would have to cross-reference them to understand the difference.

terracotta/drivers/base_classes.py
```diff
@@ -0,0 +1,135 @@
...
+        pass
+
+    @abstractmethod
+    def get_metadata(self, keys: KeysType) -> Optional[Dict[str, Any]]:
+        """Return all stored metadata for given keys."""
+        pass
+
+    @abstractmethod
+    def insert(self, keys: KeysType,
+               path: Any, **kwargs: Any) -> None:
```
nickeopti, line 98: Is there a reason to have the `**kwargs` here? If not, then let's remove it

mrpgraae: I guess it's to allow for backend-specific parameters. But actually, `TerracottaDriver` calls `MetaStore.insert()` like this
```python
self.meta_store.insert(
            keys=keys,
            path=override_path or path,
            metadata=metadata,
            **kwargs
        )
```
So in order for the abstract interface to conform to that, it should look like this
```python
@abstractmethod
def insert(self, keys: KeysType, path: Any, 
           metadata: Optional[Mapping[str, Any]], **kwargs: Any) -> None:
```

nickeopti: > I used this to indicate that derived classes may take additional keyword args.

Right, I see. That makes sense. But should we specify 
```python
    def insert(self, keys: KeysType, path: str, metadata: Mapping[str, Any] = None) -> None:
```
instead? Unless we need that extra flexibility, I would prefer to specify arguments exactly.

Note that I have actually quietly done that in `RelationalMetaStore`, which does not accept arbitrary `**kwargs` in its `insert` (and wouldn't do anything with them, in any case) -- actually making `Terracotta` driver fail if you give additional arguments to its `insert` when using a `RelationalMetaStore`. I've done the same for `compute_metadata` in  `RasterStore` and `TerracottaDriver`.

nickeopti: > So in order for the abstract interface to conform to that, it should look like this

Basically, I'm advocating that we shall do this the other way around, unless we actually need that flexibility. If possible, which I believe it is here, I think it's better to design the interfaces to describe our usage for them, rather than making them take arbitrary `**kwargs`, which subsequently requires us to be less specific down the line.

dionhaefner: Fine with me.

nickeopti: Can we image `insert`ing things besides these
```python
def insert(
        self,
        keys: KeysType,
        path: str, *,
        metadata: Mapping[str, Any] = None
    ) -> None
```
?
If so, this will require a slight modification to `RelationalMetaStore.insert`. If not, then I prefer to specify the arguments.

dionhaefner: As long as all driver combinations have the same signature I agree that we should be explicit about the arguments.

nickeopti: All relational drivers does, at least 😄 I'll specify it. We can always revert if we ever encounter a situation where we want to insert other things

nickeopti: Can I specify `path` from `Any` to `str` as well?

dionhaefner: Yes

dionhaefner:
Great job everyone - a real team effort.

PR #252, opened 2022-02-11
Test on python 3.10

### LINKED ISSUE #251: Import error on `from collections import Mapping` when running on Python 3.10
When installing from source using `conda env create -f environment.yml`, I get Python 3.10 installed due to the dependency specifier `python>=3.6`.

However, running the server and querying `/singleband` results in errors with message `AttributeError: module 'collections' has no attribute 'MutableMapping'`

Probably related: https://stackoverflow.com/questions/70195545/module-collections-has-no-attribute-mapping-issue-on-macos-for-sdk-installa

Manually editing `environment.yml` to specify `python==3.9` fixed the problem for me.

Suggestion: move from `from collections import Mapping` to `from collections.abc import Mapping`, or mark Python 3.10 as not supported yet

dionhaefner:
Fixes #251


dionhaefner:
Failing tests will be fixed by #240.

PR #253, opened 2022-02-18
Update ingest.py example after changing name from create-database

hummeltech:
This was mostly taken care of [here](https://github.com/DHI-GRAS/terracotta/commit/5aba95fd22c1060b72955967f17987e251bfd4b3) & [here](https://github.com/DHI-GRAS/terracotta/commit/dc31025507af4d683c9733507304a38ae623a0b3)


dionhaefner:
Good catch, thanks!

PR #256, opened 2022-02-28
Add writable flag to meta stores

### LINKED ISSUE #255: TerracottaDriver.get_metadata assumes meta store is writeable when lazy-loading metadata
When lazy-loading metadata, `TerracottaDriver.get_metadata` assumes that the `meta_store` has an implemented `insert` method:
```python
metadata = self.meta_store.get_metadata(keys)

if metadata is None:
    # metadata is not computed yet, trigger lazy loading
    dataset = self.get_datasets(keys)
    if not dataset:
        raise exceptions.DatasetNotFoundError('No dataset found')

    path = squeeze(dataset.values())
    metadata = self.compute_metadata(path, max_shape=self.LAZY_LOADING_MAX_SHAPE)
    self.insert(keys, path, metadata=metadata)

    # ensure standardized/consistent output (types and floating point precision)
    metadata = self.meta_store.get_metadata(keys)
    assert metadata is not None

return metadata
```

However, this behaviour fails when using `RemoteSQLiteMetaStore`, as that meta store does not implement insertions. As far as I can tell, this problem also existed before the SQLAlchemy refactor (note how `insert` is overwritten to not be implemented here https://github.com/DHI-GRAS/terracotta/blob/30730e78e204b573d1c3eb755a0107e3f73021f0/terracotta/drivers/sqlite_remote.py#L139, while `get_metadata` assumes `insert` works here https://github.com/DHI-GRAS/terracotta/blob/30730e78e204b573d1c3eb755a0107e3f73021f0/terracotta/drivers/sqlite.py#L330).

@kiksekage encountered this problem while developing a STAC driver, which will also not support insertions.
We see two possible solutions:

a) Wrap the insertion into a `try-except` block:
```python
metadata = self.compute_metadata(path, max_shape=self.LAZY_LOADING_MAX_SHAPE)

try:
    self.insert(keys, path, metadata=metadata)
    # ensure standardized/consistent output (types and floating point precision)
    metadata = self.meta_store.get_metadata(keys)
    assert metadata is not None
except NotImplementedError:
    pass
```

b) Introduce a `PERSIST_LAZY_LOADED_METADATA` setting (which defaults to `True`):
```python
metadata = self.compute_metadata(path, max_shape=self.LAZY_LOADING_MAX_SHAPE)

if self.PERSIST_LAZY_LOADED_METADATA:
    self.insert(keys, path, metadata=metadata)
    # ensure standardized/consistent output (types and floating point precision)
    metadata = self.meta_store.get_metadata(keys)
    assert metadata is not None
```

What do you think?

kiksekage:
Fixes #255

The `MetaStore` class is set to be writable by default:
https://github.com/DHI-GRAS/terracotta/blob/a3aedcbfcd95aed279dd9754434b6409462b4bc1/terracotta/drivers/base_classes.py#L41

Added a very thin decorator to throw the new `DatabaseNotWritable` exception:
https://github.com/DHI-GRAS/terracotta/blob/a3aedcbfcd95aed279dd9754434b6409462b4bc1/terracotta/drivers/terracotta_driver.py#L23-L34

And then we access the flag directly when getting metadata from the meta store:
https://github.com/DHI-GRAS/terracotta/blob/22353ca9c3f99dbe4cdb70b7497d93162827713a/terracotta/drivers/terracotta_driver.py#L187-L194

Me and @nickeopti were discussing whether or not it was cleanest to handle the database "writability" in `TerracottaDriver` or all meta stores individually.

### DIFF
--- a/terracotta/drivers/sqlite_remote_meta_store.py
+++ b/terracotta/drivers/sqlite_remote_meta_store.py
@@ -70,7 +70,7 @@ class RemoteSQLiteMetaStore(SQLiteMetaStore):
     Warning:
 
         This driver is read-only. Any attempts to use the create, insert, or delete methods
-        will throw a DatabaseNotWritable.
+        will throw a DatabaseNotWritableError.
 
     """
 
@@ -135,6 +135,7 @@ class RemoteSQLiteMetaStore(SQLiteMetaStore):
         self._update_db(self._remote_path, self._local_path)
         super()._connection_callback()
 
+    # Always raises DatabaseNotWritableError
     @requires_writable
     def create(self, *args: Any, **kwargs: Any) -> None:
         pass
--- a/terracotta/drivers/sqlite_remote_meta_store.py
+++ b/terracotta/drivers/sqlite_remote_meta_store.py
@@ -73,6 +73,8 @@ class RemoteSQLiteMetaStore(SQLiteMetaStore):
 
     """
 
+    WRITEABLE: bool = False
+
     def __init__(self, remote_path: Union[str, Path]) -> None:
         """Initialize the RemoteSQLiteDriver.
 
--- a/terracotta/drivers/sqlite_remote_meta_store.py
+++ b/terracotta/drivers/sqlite_remote_meta_store.py
@@ -11,11 +11,10 @@ import tempfile
 import time
 import urllib.parse as urlparse
 from pathlib import Path
-from typing import Any, Iterator, Union
+from typing import Iterator, Union
 
 from terracotta import exceptions, get_settings
 from terracotta.drivers.sqlite_meta_store import SQLiteMetaStore
-from terracotta.drivers.base_classes import requires_writable
 from terracotta.profile import trace
 
 logger = logging.getLogger(__name__)
@@ -134,19 +133,6 @@ class RemoteSQLiteMetaStore(SQLiteMetaStore):
     def _connection_callback(self) -> None:
         self._update_db(self._remote_path, self._local_path)
         super()._connection_callback()
-
-    # Always raises DatabaseNotWritableError
-    @requires_writable
-    def create(self, *args: Any, **kwargs: Any) -> None:
-        pass
-
-    @requires_writable
-    def insert(self, *args: Any, **kwargs: Any) -> None:
-        pass
-
-    @requires_writable
-    def delete(self, *args: Any, **kwargs: Any) -> None:
-        pass
 
     def __del__(self) -> None:
         """Clean up temporary database upon exit"""
--- a/terracotta/drivers/sqlite_remote_meta_store.py
+++ b/terracotta/drivers/sqlite_remote_meta_store.py
@@ -15,6 +15,7 @@ from typing import Any, Iterator, Union
 
 from terracotta import exceptions, get_settings
 from terracotta.drivers.sqlite_meta_store import SQLiteMetaStore
+from terracotta.drivers.base_classes import requires_writable
 from terracotta.profile import trace
 
 logger = logging.getLogger(__name__)
@@ -69,11 +70,11 @@ class RemoteSQLiteMetaStore(SQLiteMetaStore):
     Warning:
 
         This driver is read-only. Any attempts to use the create, insert, or delete methods
-        will throw a NotImplementedError.
+        will throw a DatabaseNotWritable.
 
     """
 
-    WRITEABLE: bool = False
+    _WRITABLE: bool = False
 
     def __init__(self, remote_path: Union[str, Path]) -> None:
         """Initialize the RemoteSQLiteDriver.
@@ -134,14 +135,17 @@ class RemoteSQLiteMetaStore(SQLiteMetaStore):
         self._update_db(self._remote_path, self._local_path)
         super()._connection_callback()
 
+    @requires_writable
     def create(self, *args: Any, **kwargs: Any) -> None:
-        raise NotImplementedError('Remote SQLite databases are read-only')
+        pass
 
+    @requires_writable
     def insert(self, *args: Any, **kwargs: Any) -> None:
-        raise NotImplementedError('Remote SQLite databases are read-only')
+        pass
 
+    @requires_writable
     def delete(self, *args: Any, **kwargs: Any) -> None:
-        raise NotImplementedError('Remote SQLite databases are read-only')
+        pass
 
     def __del__(self) -> None:
         """Clean up temporary database upon exit"""
--- a/terracotta/drivers/terracotta_driver.py
+++ b/terracotta/drivers/terracotta_driver.py
@@ -5,8 +5,7 @@ The driver to interact with.
 
 import contextlib
 from collections import OrderedDict
-import functools
-from typing import (Any, Callable, Collection, Dict, List, Mapping,
+from typing import (Any, Collection, Dict, List, Mapping,
                     Optional, Sequence, Tuple, TypeVar, Union)
 
 import terracotta
@@ -18,17 +17,6 @@ from terracotta.drivers.base_classes import (KeysType, MetaStore,
 ExtendedKeysType = Union[Sequence[str], Mapping[str, str]]
 ExtendedMultiValueKeysType = Union[Sequence[str], Mapping[str, Union[str, List[str]]]]
 T = TypeVar('T')
-
-
-def requires_writable(fun: Callable[..., T]) -> Callable[..., T]:
-    @functools.wraps(fun)
-    def inner(self: "TerracottaDriver", *args: Any, **kwargs: Any) -> T:
-        if self.meta_store.WRITABLE:
-            return fun(self, *args, **kwargs)
-        else:
-            raise exceptions.DatabaseNotWritable("Database not writable")
-
-    return inner
 
 
 def squeeze(iterable: Collection[T]) -> T:
@@ -70,7 +58,6 @@ class TerracottaDriver:
         """
         return self.meta_store.key_names
 
-    @requires_writable
     def create(self, keys: Sequence[str], *,
                key_descriptions: Mapping[str, str] = None) -> None:
         """Create a new, empty metadata store.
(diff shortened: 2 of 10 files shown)


dionhaefner:
I like introducing a new exception. But I think it would be cleaner if the raster drivers themselves would check for writability and raise the exception if necessary, and we can just let that bubble up to the client. Then we can also have this:

```python
class RemoteSQLiteRasterStore:
    @requires_writable
    def insert(self, *args, **kwargs):
        pass  # always unwritable
```

In this case `writable` could be a private attribute `self._writable`, and all interactions could happen via this exception. For example, to raise a better error in conjunction with lazy loading we could catch the exception and re-raise it with a different message:

```python
try:
    self.insert(keys, path, metadata=metadata) 
except DatabaseNotWritable as exc:
    raise DatabaseNotWritable("Lazy loading requires a writable database") from exc
```

kiksekage:
I moved the handling into the meta stores instead, with the explicit catching as well. If it looks good to you @dionhaefner I'll write some tests to up the coverage 👍

terracotta/drivers/sqlite_remote_meta_store.py
```diff
@@ -132,14 +135,17 @@ def _connection_callback(self) -> None:
         self._update_db(self._remote_path, self._local_path)
         super()._connection_callback()
 
+    @requires_writable
```
dionhaefner, line 138: Please add a comment that this always raises

terracotta/drivers/terracotta_driver.py
```diff
@@ -169,11 +169,17 @@ def get_metadata(self, keys: ExtendedKeysType) -> Dict[str, Any]:
 
             path = squeeze(dataset.values())
             metadata = self.compute_metadata(path, max_shape=self.LAZY_LOADING_MAX_SHAPE)
-            self.insert(keys, path, metadata=metadata)
 
-            # ensure standardized/consistent output (types and floating point precision)
-            metadata = self.meta_store.get_metadata(keys)
-            assert metadata is not None
+            try:
+                self.insert(keys, path, metadata=metadata)
```
dionhaefner, line 174: Only this line should be in `try-except`

terracotta/exceptions.py
```diff
@@ -24,5 +24,9 @@ class InvalidDatabaseError(Exception):
     pass
 
 
+class DatabaseNotWritable(Exception):
```
dionhaefner, line 27: Should be `DatabaseNotWritableError`

dionhaefner (commented):
Yes that looks pretty good. Thanks!

I think we will also need an error handler for the new exception but the tests should reveal that.

BTW, can we check whether the other drivers are writable via SQLalchemy? It should just boil down to checking whether the current user has write permissions on the used database.

kiksekage:
The failing macos tests go through now and the PR should be ready to merge if you have the time @dionhaefner @mrpgraae 👍

dionhaefner:
> I think we will also need an error handler for the new exception but the tests should reveal that.

> BTW, can we check whether the other drivers are writable via SQLalchemy? It should just boil down to checking whether the current user has write permissions on the used database.

Could you follow up on these points? What's the status?

kiksekage:
> Could you follow up on these points? What's the status?

Good point, I will have a look at that.

kiksekage:
> BTW, can we check whether the other drivers are writable via SQLalchemy? It should just boil down to checking whether the current user has write permissions on the used database.

I can't seem to find any ORM/core way of doing this in sqlalchemy, though I might be looking in the wrong directions. [This](https://github.com/sqlalchemy/sqlalchemy/issues/4239) and [this](https://github.com/sqlalchemy/sqlalchemy/issues/3326) (very old) issue seem to indicate that interfacing with user privileges is done the raw SQL query way, and then we are back to having to discern between which backend we are using and what syntax to use.
I guess we could do a write to any/all of the four tables and a rollback at `meta_store` instantiation time, but that also doesn't sit right with me.

> I think we will also need an error handler for the new exception but the tests should reveal that.

I confused myself, does this refer to adding `exceptions.DatabaseNotWritableError` to the API error handlers here: https://github.com/DHI-GRAS/terracotta/blob/b5ee523be89dba5a622633751f4181ad7f827246/terracotta/server/flask_api.py#L41

dionhaefner:
Too bad with the permissions. Maybe doing it in SQL is not so bad? But then again it might not be portable between different DB vendors? Your call.

> I confused myself, does this refer to adding exceptions.DatabaseNotWritableError to the API error handlers here: 

Yes, exactly. Right now this will throw a 500 internal server error, but maybe we would want something more descriptive.

nickeopti (commented):
> > I think we will also need an error handler for the new exception but the tests should reveal that.

> I confused myself, does this refer to adding exceptions.DatabaseNotWritableError to the API error handlers here: 

Yeah, that will need a flask error handler, like the ones you referenced, @kiksekage. However, it seems that we don't have any tests for lazy-loading metadata on API/server level. I guess we should add that. I must admit I haven't quite grasped how that setup works, so I haven't added one.

terracotta/drivers/terracotta_driver.py
```diff
@@ -169,7 +169,13 @@ def get_metadata(self, keys: ExtendedKeysType) -> Dict[str, Any]:
...
             path = squeeze(dataset.values())
             metadata = self.compute_metadata(path, max_shape=self.LAZY_LOADING_MAX_SHAPE)
-            self.insert(keys, path, metadata=metadata)
+
+            try:
+                self.insert(keys, path, metadata=metadata)
+            except exceptions.DatabaseNotWritableError as exc:
+                raise exceptions.DatabaseNotWritableError(
+                    "Lazy loading requires a writable database"
+                ) from exc
```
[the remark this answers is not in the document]
dionhaefner: Maybe, if you have a convincing use case. I don't really want to encourage lazy loading too much, especially not for production contexts. So the functionality is limited by design.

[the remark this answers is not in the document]
kiksekage: > And also, @kiksekage, the STAC driver you were working on, required lazy loading without writable, right?

Yeah, the choice was either a) maintain a local DB for metadata calculation results or b) only do on-the-fly metadata calculations, and b) seemed nicer by simply not having to maintain a local DB. This might still change though! 
I like the idea of a warning, and currently I guess it would only be raised when lazy loading on a sqlite-remote metastore. Since that didn't work actually work in the codebase before, it might be a small enough niche that it shouldn't affect too many users?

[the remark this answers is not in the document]
dionhaefner: I suggest we keep it as it is for now and discuss this in a separate issue.

terracotta/drivers/sqlite_remote_meta_store.py
```diff
@@ -132,14 +135,18 @@ def _connection_callback(self) -> None:
...
     def insert(self, *args: Any, **kwargs: Any) -> None:
-        raise NotImplementedError('Remote SQLite databases are read-only')
+        pass
 
+    @requires_writable
     def delete(self, *args: Any, **kwargs: Any) -> None:
-        raise NotImplementedError('Remote SQLite databases are read-only')
+        pass
```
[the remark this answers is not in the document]
kiksekage: Yeah I agree, simply `pass`ing doesn't help the understanding anymore than what is already conveyed.

kiksekage:
> Too bad with the permissions. Maybe doing it in SQL is not so bad? But then again it might not be portable between different DB vendors? Your call.

I think, with all the work that went into sqlachemy-fying the DB handling it hit me the wrong way to start doing some vendor-specific things again. I'm in favor of not checking it and letting it be up to the user.

I have added an error handler, and as far as I could tell, the only way for a API user to hit this error would be to request metadata for a dataset which is present but doesn't have any pre-calculated metadata. The test uses the already existing fixtures for generating `sqlite-remote` databases for testing so I moved that functionality to `conftest.py`.

dionhaefner:
> I think, with all the work that went into sqlachemy-fying the DB handling it hit me the wrong way to start doing some vendor-specific things again. I'm in favor of not checking it and letting it be up to the user.

Sounds like sunk cost fallacy to me :) But it's fine for now, the error from SQLalchemy should be descriptive enough to alert users as to what's going on.

tests/server/test_flask_api.py
```diff
@@ -39,6 +53,57 @@ def test_get_metadata(client, use_testdb):
...
+            f'/compute/some/value/{z}/{x}/{y}.png'
+            '?expression=v1*v2&v1=val22&v2=val23'
+            '&stretch_range=[10000,0]'
+        )
+
+    with pytest.raises(marshmallow.ValidationError):
+        debug_client.get(
+            f'/compute/some/value/{z}/{x}/{y}.png'
+            '?stretch_range=[10000,0]'
+        )
```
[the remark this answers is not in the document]
kiksekage: Makes sense! Will do.

tests/conftest.py
```diff
@@ -350,43 +345,19 @@ def use_testdb(testdb, monkeypatch):
...
+    terracotta.update_settings(DRIVER_PATH=str(testdb))
 
-    return _s3_db_factory
+    driver = terracotta.get_driver(testdb)
+    with driver.connect():
+        driver.insert(('first', 'second', 'third'), str(raster_file), skip_metadata=True)
 
+    driver.meta_store._WRITABLE = False
+    yield
+    driver.meta_store._WRITABLE = True
```
dionhaefner, line 358: You already give monkeypatch as a fixture, so why not use it? :smile: 

```suggestion
    with monkeypatch.context() as m:
        m.setattr(driver.meta_store, "_WRITABLE", False)
        yield
```

(no need to set it back)

kiksekage: Huh, TIL! :clap:

tests/drivers/test_raster_drivers.py
```diff
@@ -144,6 +146,22 @@ def test_lazy_loading(driver_path, provider, raster_file):
...
+    keys = ('some', 'keynames')
+
+    db.create(keys)
+    db.insert(['some', 'value'], str(raster_file), skip_metadata=True)
+
+    # Manually set the meta store to un-writable
+    db.meta_store._WRITABLE = False
+
+    with pytest.raises(exceptions.DatabaseNotWritableError):
+        db.get_metadata(['some', 'value'])
```
dionhaefner, line 162: I think this will need a monkeypatch context, too.

kiksekage: True, added it now.

dionhaefner (approved):
LGTM! Thanks @kiksekage

kiksekage:
> LGTM! Thanks @kiksekage

Should I do the honors of merging or will you @dionhaefner? 😄

PR #257, opened 2022-03-01
PostgreSQL Support

mrpgraae:
Adds a PostgreSQL meta store.

Would have loved to use the pure-python psycopg3, but it will unfortunately [not be supported by SQLAlchemy until 2.0](https://github.com/sqlalchemy/sqlalchemy/issues/6842#issuecomment-995211004).


terracotta/drivers/postgresql_meta_store.py
```diff
@@ -0,0 +1,80 @@
...
+    def _normalize_path(cls, path: str) -> str:
+        url = cls._parse_path(path)
+
+        path = f'{url.drivername}://{url.host}:{url.port or cls.DEFAULT_PORT}/{url.database}'
+        path = path.rstrip('/')
+        return path
+
+    def _create_database(self) -> None:
+        engine = sqla.create_engine(
+            self.url.set(database='postgres'),  # `.set()` returns a copy with changed parameters
```
mrpgraae, line 64: Postgres does not allow to connect without connecting to a specific database.
`postgres` is a database that is automatically created when a PostgreSQL cluster is created.
The Postgres docs say that this database "is meant as a default database for users and applications to connect to", so it should be a safe default.

terracotta/config.py
```diff
@@ -73,6 +73,12 @@ class TerracottaSettings(NamedTuple):
     #: MySQL database password (if not given in driver path)
     MYSQL_PASSWORD: Optional[str] = None
 
+    #: PostgreSQL database username (if not given in driver path)
+    POSTGRESQL_USER: Optional[str] = None
+
+    #: PostgreSQL database password (if not given in driver path)
+    POSTGRESQL_PASSWORD: Optional[str] = None
```
mrpgraae, line 80: We might want to rename these to `SQL_USER`/`SQL_PASSWORD` and then deprecate the `MYSQL_` fields?

mrpgraae:
Docs still need to be updated, but code should be ready for review :slightly_smiling_face:

dionhaefner:
I'm up to my neck with thesis work at the moment, so I'll trust your best judgment. I agree on deprecating the `MYSQL_` settings, having different settings just seems silly.

mrpgraae:
> I'm up to my neck with thesis work at the moment, so I'll trust your best judgment. I agree on deprecating the `MYSQL_` settings, having different settings just seems silly.

Cool, thanks Dion. I'll merge this PR and update docs separately. Hope the remaining thesis work goes smoothly 🙂

terracotta/drivers/postgresql_meta_store.py
nickeopti: That is also what I had in mind.

Perhaps add a `DEFAULT_DB_FOR_MANAGING = 'postgres'` (if you can think of a better name, please use that) class variable, such that if someone somehow have managed to delete/rename their `postgres` database, they still have the option of creating a TC database, by overwriting that class variable

nickeopti: With a brief explaining comment, I think that is good

terracotta/config.py
mrpgraae: @chapmanjacobd I have thought about this, but I think it would be best to avoid using the `PG*` variables. Every other env var that Terracotta reads is prefixed with `TC_` so for consistency I think we should stick to that.

mrpgraae: @nickeopti Okay, I won't add them here then. If we rename the existing variables, it will break backwards compatibility for users who are already using MySQL. Since we are heading for a 1.0 release, maybe that's fine?

tests/conftest.py
```diff
@@ -372,58 +381,80 @@ def test_server(testdb):
...
+                try:
+                    if provider == 'postgresql':
+                        # Postgres refuses to drop DB if any sessions are hanging around
+                        cursor.execute("SELECT pg_terminate_backend(pg_stat_activity.pid) "
+                                       "FROM pg_stat_activity "
+                                       f"WHERE pg_stat_activity.datname = '{dbpath}'")
+                    cursor.execute(f'DROP DATABASE IF EXISTS {dbpath}')
+                except driver.Warning:
+                    pass
+            connection.close()
```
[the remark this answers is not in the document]
mrpgraae: Yeah this part is messy.

Yes if that works it would definitely be a good way to get rid of some of the branching. I'll try it out.

terracotta/config.py
nickeopti: I'll leave decisions regarding backwards compatibility and deprecation behaviour to @dionhaefner, @j08lue and/or you @mrpgraae. I don't know the use cases well enough

nickeopti: See #258

tests/conftest.py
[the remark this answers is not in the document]
nickeopti: Forgot to add: use `driver.connect(verify=False)`

mrpgraae:
@nickeopti I made the requested changes now. Should be good to go I think :slightly_smiling_face:

PR #258, opened 2022-03-09
Re-add database user and password as settings, and rename from MYSQL_* to SQL_*

nickeopti:
The `SQLAlchemy` broke setting `MYSQL_USER` and `MYSQL_PASSWORD` settings. This PR fixes that.

It also renames the fields from `MYSQL_*` to the more generic `SQL_*`, in anticipation of the coming `PostgreSQL` driver (and potential future others). This is currently just a plain rename, thus breaking compatibility with deployments using the previous fields. We need to decide whether that is okay, or whether there should be some deprecation warning system first.

Ideally, we also ought to make some tests for this. I have apparently previously broken this, without noticing. That shouldn't be possible in the future. Feel free to just add such tests -- it may take a while before I'm back to do it.

### DIFF
--- a/terracotta/config.py
+++ b/terracotta/config.py
@@ -11,6 +11,8 @@ import warnings
 
 from marshmallow import Schema, fields, validate, pre_load, post_load, ValidationError
 
+from terracotta import exceptions
+
 
 class TerracottaSettings(NamedTuple):
     """Contains all non-deprecated settings for the current Terracotta instance."""
@@ -74,24 +76,23 @@ class TerracottaSettings(NamedTuple):
     #: SQL database password (if not given in driver path)
     SQL_PASSWORD: Optional[str] = None
 
+    #: Deprecated, use SQL_USER. MySQL database username (if not given in driver path)
+    MYSQL_USER: Optional[str] = None
+
+    #: Deprecated, use SQL_PASSWORD. MySQL database password (if not given in driver path)
+    MYSQL_PASSWORD: Optional[str] = None
+
+    #: Deprecated, use SQL_USER. PostgreSQL database username (if not given in driver path)
+    POSTGRESQL_USER: Optional[str] = None
+
+    #: Deprecated, use SQL_PASSWORD. PostgreSQL database password (if not given in driver path)
+    POSTGRESQL_PASSWORD: Optional[str] = None
+
     #: Use a process pool for band retrieval in parallel
     USE_MULTIPROCESSING: bool = True
 
 
-class DeprecatedTerracottaSettings(NamedTuple):
-    """Contains all deprecated, but still existing, settings."""
-    #: MySQL database username (if not given in driver path)
-    MYSQL_USER: Optional[str] = None
-
-    #: MySQL database password (if not given in driver path)
-    MYSQL_PASSWORD: Optional[str] = None
-
-    #: PostgreSQL database username (if not given in driver path)
-    POSTGRESQL_USER: Optional[str] = None
-
-    #: PostgreSQL database password (if not given in driver path)
-    POSTGRESQL_PASSWORD: Optional[str] = None
-
+AVAILABLE_SETTINGS: Tuple[str, ...] = TerracottaSettings._fields
 
 DEPRECATION_MAP: Dict[str, str] = {
     'MYSQL_USER': 'SQL_USER',
@@ -99,11 +100,6 @@ DEPRECATION_MAP: Dict[str, str] = {
     'POSTGRESQL_USER': 'SQL_USER',
     'POSTGRESQL_PASSWORD': 'SQL_PASSWORD',
 }
-
-AVAILABLE_SETTINGS: Tuple[str, ...] = (
-    *TerracottaSettings._fields,
-    *DeprecatedTerracottaSettings._fields
-)
 
 
 def _is_writable(path: str) -> bool:
@@ -149,8 +145,13 @@ class SettingSchema(Schema):
     ALLOWED_ORIGINS_METADATA = fields.List(fields.String())
     ALLOWED_ORIGINS_TILES = fields.List(fields.String())
 
-    SQL_USER = fields.String()
-    SQL_PASSWORD = fields.String()
+    SQL_USER = fields.String(allow_none=True)
+    SQL_PASSWORD = fields.String(allow_none=True)
+
+    MYSQL_USER = fields.String(allow_none=True)
+    MYSQL_PASSWORD = fields.String(allow_none=True)
+    POSTGRESQL_USER = fields.String(allow_none=True)
+    POSTGRESQL_PASSWORD = fields.String(allow_none=True)
 
     USE_MULTIPROCESSING = fields.Boolean()
 
@@ -166,6 +167,26 @@ class SettingSchema(Schema):
                     raise ValidationError(
                         f'Could not parse value for key {var} as JSON: "{val}"'
                     ) from exc
+        return data
+    
+    @pre_load
+    def handle_deprecated_fields(self, data: Dict[str, Any], **kwargs: Any) -> Dict[str, Any]:
+        for deprecated_field, new_field in DEPRECATION_MAP.items():
+            if data.get(deprecated_field):
+                if not data.get(new_field):
+                    warnings.warn(
+                        f'Setting TC_{deprecated_field} is being deprecated. '
+                        f'Please use TC_{new_field} instead.',
+                        exceptions.DeprecationWarning
+                    )
+                    data[new_field] = data[deprecated_field]
+                else:
+                    warnings.warn(
+                        f'Both the deprecated TC_{deprecated_field} setting '
+                        f'and its replacement TC_{new_field} is set: '
+                        'This may result in unexpected behaviour.',
+                        exceptions.DeprecationWarning
+                    )
         return data
 
     @post_load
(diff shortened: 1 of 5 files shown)


mrpgraae:
IMO we should just deprecate the MYSQL_ vars and make the breaking change. We'll want to do that before 1.0 in any case.

If we want to be nice, we can detect `MYSQL_` and fail with an error message saying that this has changed to `SQL_`. Then remove the error message when we release 1.0.

dionhaefner:
I would prefer a proper deprecation cycle with a warning and then remove it in the next major version. Emitting this warning should be pretty painless. I suspect that we have a good amount of users by now (judging from the traffic and downstream projects), and I don't see any reason to break things unless we need to.

mrpgraae:
> I would prefer a proper deprecation cycle with a warning and then remove it in the next major version. Emitting this warning should be pretty painless. I suspect that we have a good amount of users by now (judging from the traffic and downstream projects), and I don't see any reason to break things unless we need to.

In that case, @nickeopti ignore my approved review and let's add this warning.

mrpgraae:
> It was my understanding that DB migration is a hard requirement before the next release.

Okay, I guess I thought that we would support from the next release and onwards and not from the current release to the next.

nickeopti:
I've implemented a deprecation system for the settings. Perhaps a bit over-engineered, but I think it does the job.

One question, though, is whether we shall make the warnings more visible. Apparently the default settings in Python hide `DeprecrationWarning`s. Is that acceptable, or shall we either use something else or change the filtering levels with something like
```python
warnings.simplefilter('always', DeprecationWarning)
```
There's also a question of `PendingDeprecationWarning` versus `DeprecationWarning`. Essentially, I don't know the best practices regarding deprecation cycles.

I've not added tests for this yet. I'd like to know that the functionality is correct before adding those.

dionhaefner:
I would have approached this a bit differently. Right now this seems weird to me:

```python
>>> settings = get_settings()
>>> settings.MYSQL_USER = "foo"
>>> settings.MYSQL_USER
AttributeError: 'TerracottaSettings' object has no attribute 'MYSQL_USER'
```

I would suggest something simpler:

- Leave the deprecated settings as attributes of `TerracottaSettings`
- Use your deprecation map in `pre_load` and `post_load` to emit the warning and ensure that the new setting is written

> One question, though, is whether we shall make the warnings more visible. Apparently the default settings in Python hide DeprecrationWarnings.

IMO Python's deprecation warnings are only appropriate for *libraries*, where you don't want end users to be bothered by them (i.e., where they are only relevant to developers). This is not the case in end-user applications and we should use a different warning that is visible by default. I suggest introducing a new warning class in `exceptions.py` and using that.

nickeopti:
> I would have approached this a bit differently. Right now this seems weird to me:
> ```python
> >>> settings = get_settings()
> >>> settings.MYSQL_USER = "foo"
> >>> settings.MYSQL_USER
> AttributeError: 'TerracottaSettings' object has no attribute 'MYSQL_USER'
> ```

But is that really a concern, as `NamedTuple`s are immutable? So `settings.MYSQL_USER = "foo"` has never been possible. However, I do see that this code does not currently work with the `update_settings`, which is certainly a bug on my part.

> I would suggest something simpler:
> 
>  *   Leave the deprecated settings as attributes of TerracottaSettings
>  *   Use your deprecation map in pre_load and post_load to emit the warning and ensure that the new setting is written

That makes sense


>  >   One question, though, is whether we shall make the warnings more visible. Apparently the default settings in Python hide DeprecrationWarnings.
> 
> IMO Python's deprecation warnings are only appropriate for libraries, where you don't want end users to be bothered by them (i.e., where they are only relevant to developers). This is not the case in end-user applications and we should use a different warning that is visible by default. I suggest introducing a new warning class in exceptions.py and using that.

I'll do that.

Another question: 
I'm tempted to use a `pydantic` `BaseModel` for the settings, as that would allow us to specify the fields as well as validation more compactly, with less repetition of field names, in a single class. I just made a quick mockup, and it seems to roughly halve the number of lines of code, while arguably be easier to understand. Downside being adding a package as well as pydantic models being larger than `namedtuple`s. What is your opinion on that change?

nickeopti:
Also, when setting a deprecated setting, eg `MYSQL_USER`, do we expect:
1. that just `SQL_USER` gets set with that value, or
2. that both `MYSQL_USER` and `SQL_USER` gets set with that value?

dionhaefner:
> But is that really a concern, as NamedTuples are immutable? So settings.MYSQL_USER = "foo" has never been possible. 

Right, my bad. It should have been `update_settings` and then reading it back.

> I'm tempted to use a pydantic BaseModel for the settings, as that would allow us to specify the fields as well as validation more compactly, with less repetition of field names, in a single class. I just made a quick mockup, and it seems to roughly halve the number of lines of code, while arguably be easier to understand. Downside being adding a package as well as pydantic models being larger than namedtuples. What is your opinion on that change?

I'm not a big fan of introducing a dependency for functionality that is already present and works fine. IMO we have nothing to gain from switching at this point, but if others are positive to it I won't veto it :)

> Also, when setting a deprecated setting, eg MYSQL_USER, do we expect:
>
>    1. that just SQL_USER gets set with that value, or
>    2. that both MYSQL_USER and SQL_USER gets set with that value?

I'd go with number 2.

nickeopti:
I believe it is ready for review @dionhaefner

terracotta/config.py
```diff
@@ -150,6 +169,26 @@ def decode_lists(self, data: Dict[str, Any], **kwargs: Any) -> Dict[str, Any]:
                     ) from exc
         return data
 
+    @pre_load
+    def handle_deprecated_fields(self, data: Dict[str, Any], **kwargs: Any) -> Dict[str, Any]:
+        for deprecated_field, new_field in DEPRECATION_MAP.items():
+            if data.get(deprecated_field):
+                if not data.get(new_field):
+                    warnings.warn(
+                        f'Setting TC_{deprecated_field} is being deprecated. '
```
dionhaefner, line 178:
```suggestion
                        f'Setting TC_{deprecated_field} is deprecated and will be removed in the next major release. '
```

terracotta/config.py
```diff
@@ -67,23 +70,36 @@ class TerracottaSettings(NamedTuple):
...
     POSTGRESQL_PASSWORD: Optional[str] = None
 
     #: Use a process pool for band retrieval in parallel
     USE_MULTIPROCESSING: bool = True
 
 
-AVAILABLE_SETTINGS: Tuple[str, ...] = tuple(TerracottaSettings._fields)
+AVAILABLE_SETTINGS: Tuple[str, ...] = TerracottaSettings._fields
+
+DEPRECATION_MAP: Dict[str, str] = {
```
dionhaefner, line 97:
```suggestion
DEPRECATION_MAP: Dict[str, str] = {
    # TODO: Remove in v0.8.0
```

tests/drivers/test_drivers.py
```diff
@@ -261,3 +261,23 @@ def test_invalid_key_types(driver_path, provider):
...
+
+
+@pytest.mark.parametrize('provider', TESTABLE_DRIVERS)
+def test_use_credentials_from_settings(driver_path, provider, monkeypatch):
+    with monkeypatch.context() as m:
+        m.setenv('TC_SQL_USER', 'foo')
+        m.setenv('TC_SQL_PASSWORD', 'bar')
+
+        from terracotta import drivers, update_settings
+        update_settings()
```
dionhaefner, line 273: This only works because these settings have not been set explicitly by another test. We should guard against this, e.g. by clearing `_overwritten_settings` at the start of the test.

tests/test_config.py
```diff
@@ -76,3 +76,26 @@ def test_update_config():
...
+
+
+def test_deprecation_behaviour(monkeypatch):
+    from terracotta import config, exceptions, get_settings, update_settings
+    for deprecated_field, new_field in config.DEPRECATION_MAP.items():
+        with monkeypatch.context() as m:
+            m.setenv(f'TC_{deprecated_field}', 'foo')
+
+            with pytest.warns(exceptions.DeprecationWarning) as warning:
+                update_settings()
```
dionhaefner, line 88: Same here.

terracotta/config.py
```diff
@@ -150,6 +169,26 @@ def decode_lists(self, data: Dict[str, Any], **kwargs: Any) -> Dict[str, Any]:
...
+                        exceptions.DeprecationWarning
+                    )
+                    data[new_field] = data[deprecated_field]
+                else:
+                    warnings.warn(
+                        f'Both the deprecated TC_{deprecated_field} setting '
+                        f'and its replacement TC_{new_field} is set: '
+                        'This may result in unexpected behaviour.',
+                        exceptions.DeprecationWarning
+                    )
```
dionhaefner, line 189: I would remove this warning and just raise the warning above unconditionally.

terracotta/drivers/relational_meta_store.py
```diff
@@ -115,8 +115,8 @@ def _parse_path(cls, connection_string: str) -> URL:
 
         url = URL.create(
             drivername=f'{cls.SQL_DIALECT}+{cls.SQL_DRIVER}',
-            username=con_params.username,
-            password=con_params.password,
+            username=con_params.username or terracotta.get_settings().SQL_USER,
+            password=con_params.password or terracotta.get_settings().SQL_PASSWORD,
```
dionhaefner, line 119: I would prefer getting the settings once.

tests/drivers/test_drivers.py
nickeopti: Or shall we either amend that fixture to perform cleanup before yielding, or create a new fixture ala `ensure_default_settings`?

dionhaefner: Right, I forgot about that. Carry on.

dionhaefner:
Thanks @nickeopti !

PR #260, opened 2022-03-29
Test using postgresql meta store as well

nickeopti:
Include the `postgresql` driver/meta store for testing in `test_raster_drivers.py`. I assume this shall be done?

We arguably ought to restructure the tests to match our new raster/meta store structure, but I'll leave that for the future 😄

### DIFF
--- a/tests/drivers/test_raster_drivers.py
+++ b/tests/drivers/test_raster_drivers.py
@@ -5,7 +5,7 @@ import time
 
 import numpy as np
 
-DRIVERS = ['sqlite', 'mysql', 'postgresql']
+DRIVERS = ['sqlite']
 METADATA_KEYS = ('bounds', 'range', 'mean', 'stdev', 'percentiles', 'metadata')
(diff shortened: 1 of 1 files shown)


tests/drivers/test_raster_drivers.py
```diff
@@ -5,7 +5,7 @@
 
 import numpy as np
 
-DRIVERS = ['sqlite', 'mysql']
+DRIVERS = ['sqlite', 'mysql', 'postgresql']
```
dionhaefner, line 8: I don't really see the point, since the meta storage is now completely decoupled from the raster retrieval. I would actually suggest this instead:

```suggestion
DRIVERS = ['sqlite']
```

mrpgraae: Yeah I agree, as long as the postgres metastore interface works as expected then it should not have any impact on the raster driver. We are doing unit tests after all.

PR #265, opened 2022-04-29
Set default db connection timeout to 100 seconds

panakouris:
I just noticed that in the last version it seems that from the time I run `tc.get_driver()` I have 10 seconds until the connection to the DB dies. I suggest that the default value should be much higher that 10 seconds. Set it to 100 for now.


dionhaefner:
This timeout should be the timeout for *establishing* a connection, not the time it's kept alive. Did the semantics change in the SQLalchemy port? @nickeopti

panakouris:
Hmmm, I think you are correct. I will dig further into it next week. But the connection errors I get are quite often at the moment. Can also be related to overused database.

PR #267, opened 2022-05-02
Issue 266

### LINKED ISSUE #266: MySQL - Lost connection to MySQL server during query
I finally managed to reproduce this issue mentioned also in #265, and found a possible solution to this. The issue is that after some period of inactivity from the time I was getting the driver with `tc.get_driver()` to the time I was interacting with the database (for example with `driver.get_datasets()`), I keep getting this error message `OperationalError: (2013, 'Lost connection to MySQL server during query')`. From a small test I performed, I saw that the period of inactivity is not always the same, most of the times between 5 and 20 seconds, which means that the reason for the dropped connection is not some hardcoded timeout.  The solution that I have tested and seems to work fine is setting `pool_pre_ping=True` when creating the sqlalchemy engine https://github.com/DHI-GRAS/terracotta/blob/2e8bcdb718bd911569248c0cd1482a29c0ae24a1/terracotta/drivers/relational_meta_store.py#L87 .
From SQLAlchemy docs `The “pre ping” feature will normally emit SQL equivalent to “SELECT 1” each time a connection is checked out from the pool; if an error is raised that is detected as a “disconnect” situation, the connection will be immediately recycled, and all other pooled connections older than the current time are invalidated, so that the next time they are checked out, they will also be recycled before use.` 
I tested up to 400 seconds of inactivity between `tc.get_driver()` and `driver.get_datasets()`, and I haven't received any error. What do you think @Nick0693 and @dionhaefner ?

panakouris:
This fixes #266.
Set `pool_pre_ping` to `True` when creating the sqlalchemy engine as discussed in #266

### DIFF
--- a/terracotta/drivers/relational_meta_store.py
+++ b/terracotta/drivers/relational_meta_store.py
@@ -89,6 +89,7 @@ class RelationalMetaStore(MetaStore, ABC):
             echo=False,
             future=True,
             connect_args={self.SQL_TIMEOUT_KEY: db_connection_timeout},
+            # automatically re-spawn stale connections, see terracotta#266
             pool_pre_ping=True
         )
         self.sqla_metadata = sqla.MetaData()
--- a/terracotta/drivers/relational_meta_store.py
+++ b/terracotta/drivers/relational_meta_store.py
@@ -88,7 +88,8 @@ class RelationalMetaStore(MetaStore, ABC):
             self.url,
             echo=False,
             future=True,
-            connect_args={self.SQL_TIMEOUT_KEY: db_connection_timeout}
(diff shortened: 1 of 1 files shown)


terracotta/drivers/relational_meta_store.py
```diff
@@ -88,7 +88,8 @@ def __init__(self, path: str) -> None:
             self.url,
             echo=False,
             future=True,
-            connect_args={self.SQL_TIMEOUT_KEY: db_connection_timeout}
+            connect_args={self.SQL_TIMEOUT_KEY: db_connection_timeout},
+            pool_pre_ping=True
```
dionhaefner, line 92:
```suggestion
            # automatically re-spawn stale connections, see terracotta#266
            pool_pre_ping=True
```

Issue #261, opened 2022-04-10
Could you help upgrade the vulnerble dependency in terracotta ? 


JoeGardner000:
Hi, @dionhaefner , @mrpgraae , I'd like to report a vulnerability issue in **terracotta_0.7.5**.
### Issue Description
I noticed that **terracotta_0.7.5** directly depends on **rasterio_1.2.10**.
However, **rasterio_1.2.10** sufferes from the vulnerabilites which the C libraries exposed as following dependency graph shows. 
### Dependency Graph between Python and Shared Libraries
![image (11)](https://user-images.githubusercontent.com/103017573/162617943-87fb3995-b1b0-49aa-887e-6c73b4f5cbb3.png)
### Suggested Vulnerability Patch Versions
**rasterio** has upgraded these vulnerable C libraries to patch versions refer to [issue url](https://github.com/rasterio/rasterio-wheels/issues/80#issue-1194815122).

Python build tools cannot report vulnerable C libraries, which may induce potential security issues to many downstream Python projects.
As a popular python package (**terracotta** has **1,074** downloads per month), could you please upgrade this vulnerable dependency?

Thanks for your help~
Best regards,
Joe Gardner

dionhaefner:
Thanks for the report.

> I noticed that terracotta_0.7.5 directly depends on rasterio_1.2.10.

Where did you get that from? We do not depend on any specific version of rasterio, the only requirement is `>=1.0`. I guess we could exclude `1.2.10` if it is the only vulnerable version, but I assume that rasterio has been using the same HDF5 version for quite a while?

Also, I see no indication that this is fixed in recent rasterio wheels. The build repo still seems to use the same version:

https://github.com/rasterio/rasterio-wheels/blob/master/env_vars.sh#L15

JoeGardner000:
@dionhaefner , thanks for your feedback.

> Also, I see no indication that this is fixed in recent rasterio wheels.

Rasterio has upgrade HDF5 to patch version in `master`, may release in the next version.

dionhaefner:
OK, let's wait for the next release then and see what happens.

dionhaefner:
Bumped rasterio to 1.3 on master, so this should be fixed.

PR #269, opened 2022-07-11
Require rasterio 1.3

dionhaefner:
Investigating #261

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ channels:
 dependencies:
   - python>=3.7
   - numpy
-  - rasterio>=1.0
+  - rasterio>=1.3.0
   - shapely
   - crick
   - pip
--- a/setup.py
+++ b/setup.py
@@ -70,7 +70,7 @@ setup(
         'pillow',
         'pyyaml>=3.10',  # downstream dependency of apispec
         'shapely',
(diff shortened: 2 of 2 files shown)


dionhaefner:
Looks like it works fine, but this tightens version requirements quite a bit (Python 3.7 to 3.8 and `rasterio` from 1.0 to 1.3). On the other hand, Python 3.10 is now fully supported.

Any opinions @mrpgraae @j08lue @nickeopti ?

mrpgraae:
I guess the concern here is that we don't know what other dependencies users might have in their ingestion/processing pipelines and that these tighter constraints might break their environments?

dionhaefner:
Yes. On the other hand there are no Python 3.10 wheels for `rasterio<1.3.0`, so the update will be forced at some point.

Other options are:

1. Bisect rasterio versions to track down which wheels use vulnerable C libraries, then exclude only those.
2. Emit a warning urging users to upgrade to rasterio>=1.3.
3. Do nothing.

mrpgraae:
>On the other hand there are no Python 3.10 wheels for rasterio<1.3.0, so the update will be forced at some point.

Okay, in that case, I think now is a fine time to enforce this.

Issue #271, opened 2022-07-26
More liberal pattern when running `terracotta ingest`


tomalrussell:
It would be convenient to allow some punctuation marks as well as alphanumeric characters in the regex pattern matching key values.

I have a raster pattern like:
-  `{type}__rp_{rp}__rcp_{rcp}__epoch_{epoch}__gcm_{gcm}.tif`
and files like:
- `cyclone__rp_10__rcp_8.5__epoch_2050__gcm_CMCC-CM2-VHR4.tif`
- `river__rp_2__rcp_8.5__epoch_2030__gcm_MIROC-ESM-CHEM.tif`

The regex match is too strict to allow the `.` or `-` in values - could it be relaxed? I can patch the filenames as a workaround, but a quick edit to the key bit of the regex so it's just `[^_]+` seems to work okay locally:
https://github.com/DHI-GRAS/terracotta/blob/b7c67c3c2736401295644c1e8882b3f0f013bb5c/terracotta/scripts/click_types.py#L74

dionhaefner:
Unfortunately this is an unsolvable problem. Raster patterns are a confusing mess, and I am hesitant to make it even more confusing by capturing more stuff.

Is there any particular reason why you don't want to use the Python API for ingestion like we recommend in the docs?

I *could* be nudged to support regex patterns as a power-user feature:

```bash
$ terracotta ingest --raster-regex "(?P<type>\w+)__rp_(?P<rp>\d+)__rcp_(?P<rcp>\d+\.\d+)__epoch_(?P<epoch>\d+)__gcm_(?P<gcm>[\w-]+)\.tif"
```

But I think you have to agree that the patterns are quite messy, so it might be easier to use the Python API :)

tomalrussell:
I can see the problem. I guess I asked because the `ingest` subcommand almost does what I want and tweaking it seemed easier than learning how to use the Python API.

I can't see myself typing out that `--raster-regex` example (correctly, first time!), and at that point I'd be writing some kind of script anyway.

I wonder if a simpler example script in the docs might help:

```python#!/usr/bin/env python3
import os
from typing import Dict, List

import tqdm
import terracotta

# Define the location of the SQLite database
# (this will be created if it doesn't already exist)
DB_NAME = f"./terracotta.sqlite"

# Define the list of keys that will be used to identify datasets.
# (these need to match the key_values dicts defined in RASTER_FILES below)
KEYS = ["type", "rp", "rcp", "epoch", "gcm"]

[...]

PR #277, opened 2022-08-23
Add a first example ingestion script to the docs

tomalrussell:
This adds the example discussed in #271 to the getting started docs, aiming to provide a first example script that uses as little as possible of the terracotta API and not much core Python or other libraries either! Hopefully a good starting point for more novice users.


dionhaefner:
Fantastic, thanks!

Issue #309, opened 2023-07-28
Reduction in /metadata API calls for multiple datasets 


DenizYil:
For a variety of projects, we've had to query `/datasets` and for each dataset, query the corresponding metadata using the `/metadata` endpoint. This often results in a lot of requests (sometimes 200+), which causes a slight delay and some extra load, that could potentially be unavoided.

We save our own data in the `metadata` column in the `metadata` table. This is primarily the data we've been interested in, but I believe there have been cases where we've been interested in the other columns (excluding keys).

Would there be a solution where these requests could be grouped into one? I'm happy to elaborate and submit a PR based on any approach that is agreed upon. This would be a beneficial addition to a multitude of internal projects.

j08lue:
For the record - we discussed this a while back (2 years ago), in a Slack thread, though (I just sent you the link). Let us make sure we share the arguments / conclusions here, for posterity.

Btw, this is somewhat related to 
- #104 

in terms of providing optimized interfaces for bulk-retrieving data to populate UI elements for dataset display / selection.

The gist of the discussion 2 years ago was

1. You may want to do lazy loading in the client instead of retrieving all the data, to avoid long initial loading times?
2. However, if you _have_ to do this server-side, perhaps you could extend Terracotta for your use case (we discussed a plugin, although we **do not have any guidance on how to develop plugins**)
3. This sparse and use-case-specific metadata retrieval sounds like a nice case for [GraphQL](https://github.com/graphql-python/graphene-sqlalchemy), but maybe that is way to big a cannon...
4. @dionhaefner recommended: 
   > From the top of my head, I would maybe make this endpoint accept a POST request with all requested keys as request body. Perhaps also restrict which entries you want, retrieving >1k geometry footprints when all you need is cloudcover would blow up your response size.

Since then, STAC has seen more uptake and could be an alternative route to explore for exposing metadata-heavy EO data collections in a programmatic way.

dionhaefner:
PRs welcome. I would probably accept the design proposed above if implemented cleanly:

> make this endpoint accept a POST request with all requested keys as request body. Perhaps also restrict which entries you want, retrieving >1k geometry footprints when all you need is cloudcover would blow up your response size.

DenizYil:
During a conversation with Jonas, I proposed introducing a `?metadata=true` flag to the `/datasets` endpoint. These are my key reasons for it:

[...]

PR #311, opened 2023-08-09
feat: add endpoint for querying multiple metadata

DenizYil:
This PR closes https://github.com/DHI-GRAS/terracotta/issues/309

This PR provides the ability for users to send a `POST` request to the `/metadata` endpoint, where they can filter the columns has path parameters, e.g. `POST /metadata/bounds`, which will only return the bounds (and the keys used, which is default on all)

Notes:
- I had a challenge with the naming convention and the structure itself. Should this endpoint be in its own file, or share a file with the `GET` request?
- Should the methods be called `post_metadata` or `get_multiple_metadata` or `get_metadatas` or something else..? For now, I named the endpoint `post_metadata` and every other method `get_multiple_metadata`.
- Should the user be able to specify e.g. `bounds_south` as a column or should we only allow them to specify `bounds`? (same with `range` & `min/max`). For now, I *have* added this functionality.
- Please let me know if you have a better way to do the specific-decoding handling instead of the added `_decode_specific_data` function.
- Please let me know as well if any variable names/types should be changed.
- What should we do in the case of lazy-loading? Should we allow the user to lazy-load and generate metadata for **everything**? For now, I excluded this functionality as it could take a very long time. Let me know if you'd like this added.

Please let me know if you have any feedback, as that is very appreciated. I'm excited to see this feature be used.

### DIFF
--- a/terracotta/handlers/metadata.py
+++ b/terracotta/handlers/metadata.py
@@ -46,9 +46,10 @@ def multiple_metadata(
     key_names = driver.key_names
 
     out = []
-    for dataset in datasets:
-        metadata = filter_metadata(driver.get_metadata(dataset), columns)
-        metadata["keys"] = OrderedDict(zip(key_names, dataset))
-        out.append(metadata)
+    with driver.connect():
+        for dataset in datasets:
+            metadata = filter_metadata(driver.get_metadata(dataset), columns)
+            metadata["keys"] = OrderedDict(zip(key_names, dataset))
+            out.append(metadata)
 
     return out
--- a/terracotta/handlers/metadata.py
+++ b/terracotta/handlers/metadata.py
@@ -10,9 +10,13 @@ from terracotta import get_settings, get_driver
 from terracotta.profile import trace
 
 
-def filter_metadata(metadata: Dict[str, Any], columns: Optional[List[str]]) -> Dict[str, Any]:
+def filter_metadata(
+    metadata: Dict[str, Any], columns: Optional[List[str]]
+) -> Dict[str, Any]:
     """Filter metadata by columns, if given"""
-    assert columns is None or len(columns) > 0, "columns must either be a non-empty list or None"
+    assert (
+        columns is None or len(columns) > 0
+    ), "columns must either be a non-empty list or None"
 
     if columns:
         metadata = {c: metadata[c] for c in columns}
@@ -21,7 +25,9 @@ def filter_metadata(metadata: Dict[str, Any], columns: Optional[List[str]]) -> D
 
 
 @trace("metadata_handler")
-def metadata(columns: Optional[List[str]], keys: Union[Sequence[str], Mapping[str, str]]) -> Dict[str, Any]:
+def metadata(
+    columns: Optional[List[str]], keys: Union[Sequence[str], Mapping[str, str]]
+) -> Dict[str, Any]:
     """Returns all metadata for a single dataset"""
     settings = get_settings()
     driver = get_driver(settings.DRIVER_PATH, provider=settings.DRIVER_PROVIDER)
@@ -31,7 +37,9 @@ def metadata(columns: Optional[List[str]], keys: Union[Sequence[str], Mapping[st
 
 
 @trace("multiple_metadata_handler")
-def multiple_metadata(columns: Optional[List[str]], datasets: List[List[str]]) -> List[Dict[str, Any]]:
+def multiple_metadata(
+    columns: Optional[List[str]], datasets: List[List[str]]
+) -> List[Dict[str, Any]]:
     """Returns all metadata for multiple datasets"""
     settings = get_settings()
     driver = get_driver(settings.DRIVER_PATH, provider=settings.DRIVER_PROVIDER)
--- a/terracotta/handlers/metadata.py
+++ b/terracotta/handlers/metadata.py
@@ -47,7 +47,7 @@ def multiple_metadata(
 
     out = []
     with driver.connect():
-        for dataset in datasets:
+        for dataset in datasets[:settings.MAX_POST_METADATA_KEYS]:
             metadata = filter_metadata(driver.get_metadata(dataset), columns)
             metadata["keys"] = OrderedDict(zip(key_names, dataset))
             out.append(metadata)
--- a/terracotta/handlers/metadata.py
+++ b/terracotta/handlers/metadata.py
@@ -47,7 +47,7 @@ def multiple_metadata(
 
     out = []
     with driver.connect():
-        for dataset in datasets[:settings.MAX_POST_METADATA_KEYS]:
+        for dataset in datasets[: settings.MAX_POST_METADATA_KEYS]:
             metadata = filter_metadata(driver.get_metadata(dataset), columns)
             metadata["keys"] = OrderedDict(zip(key_names, dataset))
             out.append(metadata)
--- a/terracotta/handlers/metadata.py
+++ b/terracotta/handlers/metadata.py
@@ -7,6 +7,7 @@ from typing import Mapping, Sequence, Dict, Any, Union, List, Optional
 from collections import OrderedDict
 
 from terracotta import get_settings, get_driver
+from terracotta.exceptions import InvalidArgumentsError
 from terracotta.profile import trace
 
 
@@ -45,9 +46,15 @@ def multiple_metadata(
     driver = get_driver(settings.DRIVER_PATH, provider=settings.DRIVER_PROVIDER)
     key_names = driver.key_names
 
+    if len(datasets) > settings.MAX_POST_METADATA_KEYS:
+        raise InvalidArgumentsError(
+            f"Maximum number of keys exceeded ({settings.MAX_POST_METADATA_KEYS}). "
+            f"This limit can be configured in the server settings."
+        )
+
     out = []
     with driver.connect():
-        for dataset in datasets[: settings.MAX_POST_METADATA_KEYS]:
+        for dataset in datasets:
             metadata = filter_metadata(driver.get_metadata(dataset), columns)
             metadata["keys"] = OrderedDict(zip(key_names, dataset))
             out.append(metadata)
--- a/terracotta/handlers/metadata.py
+++ b/terracotta/handlers/metadata.py
@@ -3,18 +3,44 @@
 Handle /metadata API endpoint.
 """
 
-from typing import Mapping, Sequence, Dict, Any, Union
+from typing import Mapping, Sequence, Dict, Any, Union, List, Optional
 from collections import OrderedDict
 
 from terracotta import get_settings, get_driver
 from terracotta.profile import trace
 
 
+def filter_metadata(metadata: Dict[str, Any], columns: Optional[List[str]]) -> Dict[str, Any]:
+    """Filter metadata by columns, if given"""
+    assert columns is None or len(columns) > 0, "columns must either be a non-empty list or None"
+
+    if columns:
+        metadata = {c: metadata[c] for c in columns}
+
+    return metadata
+
+
 @trace("metadata_handler")
-def metadata(keys: Union[Sequence[str], Mapping[str, str]]) -> Dict[str, Any]:
+def metadata(columns: Optional[List[str]], keys: Union[Sequence[str], Mapping[str, str]]) -> Dict[str, Any]:
     """Returns all metadata for a single dataset"""
     settings = get_settings()
     driver = get_driver(settings.DRIVER_PATH, provider=settings.DRIVER_PROVIDER)
-    metadata = driver.get_metadata(keys)
+    metadata = filter_metadata(driver.get_metadata(keys), columns)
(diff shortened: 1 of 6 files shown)


DenizYil:
Thanks a lot for the fast feedback @dionhaefner - it's truly appreciated. I believe I should have addressed all your changes.

Some notes:
- With that being said, I kept the `multiple_data` handler function. I think it makes sense to not have this functionality in the endpoint - but let me know if I should change this. I could merge the functionality with the existing function and add a `multi` parameter (or use `isInstance`) or something to check if the contents of the list are lists or numbers - but that feels like added complexity.

- I'll also note that on my computer (I've not run this in a real production setting yet) that performance is a lot slower. Querying 36 datasets takes 12 seconds on my computer (and it's a pretty good PC), whereas before it was almost instantaneous. I'm sure performance will be much better in production, but wanted to note it, nonetheless. 

- Let me know if you think `CommaSeparatedListField` is over-engineered. I wanted to stay true to using Marshmallow, but yeah.

Thanks a lot again, and I'm looking forward to your feedback! 😄

DenizYil:
> That seems very slow to me. Can you share the SQLite database?

Agreed.

Apologies though, perhaps I should have clarified. I'm using a PostgreSQL database which is hosted on Azure - so the DB is not running locally. I've hosted a Terracotta instance using Azure Functions, which only contains 36 datasets. This is the URL: https://100ktrees-tc.azurewebsites.net/datasets

dionhaefner:
Before we get into code review, there should be a feature to limit the amount of data to be requested (as a protection against [accidental] DOS). At first I thought we should use pagination, but then again that's probably not needed here because the number of datasets is known a-priori. How about we introduce a runtime setting for this, like `MAX_POST_METADATA_KEYS` or so, and set that to a reasonable value by default (100?).

dionhaefner:
> Apologies though, perhaps I should have clarified. I'm using a PostgreSQL database which is hosted on Azure - so the DB is not running locally. I've hosted a Terracotta instance using Azure Functions, which only contains 36 datasets. This is the URL: https://100ktrees-tc.azurewebsites.net/datasets

Can you say more about how you tested the performance of the new endpoint, then?

DenizYil:
> > Apologies though, perhaps I should have clarified. I'm using a PostgreSQL database which is hosted on Azure - so the DB is not running locally. I've hosted a Terracotta instance using Azure Functions, which only contains 36 datasets. This is the URL: https://100ktrees-tc.azurewebsites.net/datasets
> 
> Can you say more about how you tested the performance of the new endpoint, then?

I simply ran this script, and the results are consistent.
```py
import requests
import time

start = time.time()

data = requests.post(
    "http://localhost:5000/metadata?columns=[metadata]",
    json={
        "keys": [
            ["01012012", "HRL", "annually", "IMD"],
            ["01012012", "HRL", "annually", "TCD"],
            ["01012015", "HRL", "annually", "GRA"],
            ["01012015", "HRL", "annually", "IMD"],
            ["01012015", "HRL", "annually", "TCD"],
            ["01012015", "HRL", "annually", "WAW"],
            ["01012018", "HRL", "annually", "GRA"],
            ["01012018", "HRL", "annually", "IMD"],
            ["01012018", "HRL", "annually", "TCD"],
            ["01012018", "HRL", "annually", "WAW"],
            ["202206241029443", "NED", "1", "B01"],
            ["202206241029443", "NED", "1", "B02"],
            ["202206241029443", "NED", "1", "B03"],
            ["202206241029443", "NED", "1", "B04"],
            ["202206241029443", "RGB", "1", "B01"],
            ["202206241029443", "RGB", "1", "B02"],
            ["202206241029443", "RGB", "1", "B03"],
            ["202206241029443", "RGB", "1", "B04"],
            ["202206241029444", "1", "F", "P"],
            ["202206241029444", "NED", "1", "B01"],
            ["202206241029444", "NED", "1", "B02"],
            ["202206241029444", "NED", "1", "B03"],
            ["202206241029444", "NED", "1", "B04"],
            ["202206241029444", "RGB", "1", "B01"],
            ["202206241029444", "RGB", "1", "B02"],
            ["202206241029444", "RGB", "1", "B03"],
            ["202206241029444", "RGB", "1", "B04"],
            ["202304081018382", "1", "F", "P"],
            ["202304081018382", "NED", "1", "B01"],
            ["202304081018382", "NED", "1", "B02"],
            ["202304081018382", "NED", "1", "B03"],
            ["202304081018382", "NED", "1", "B04"],
            ["202304081018382", "RGB", "1", "B01"],
            ["202304081018382", "RGB", "1", "B02"],
            ["202304081018382", "RGB", "1", "B03"],
            ["202304081018382", "RGB", "1", "B04"],
        ]
    },
)

end = time.time()

print(end - start)
```

> Before we get into code review, there should be a feature to limit the amount of data to be requested (as a protection against [accidental] DOS). At first I thought we should use pagination, but then again that's probably not needed here because the number of datasets is known a-priori. How about we introduce a runtime setting for this, like `MAX_POST_METADATA_KEYS` or so, and set that to a reasonable value by default (100?).

Sure! I can add that into the PR. I think if we're going with the 1 SQL query per dataset, then 100 seems reasonable, but I do think that it seems a bit on the low-end. When this has been used in legitimate cases (like querying the `GET /metadata` endpoint), it's been in the 300s+. I think if we decide to go with the 1 SQL query per dataset, we could easily consider increasing the amount. I'll keep at 100 for now though :)

dionhaefner:
So you are running a local Terracotta server that connects to the remote PostgreSQL database?

DenizYil:
Yes. I'll do some testing tomorrow with an Azure Function that connects to an Azure PostgreSQL database to see performance in a real production environment. I'll also do some testing with a locally running database (and server). I'll keep you updated :)

terracotta/handlers/metadata.py
```diff
@@ -3,18 +3,52 @@
...
+    """Returns all metadata for multiple datasets"""
+    settings = get_settings()
+    driver = get_driver(settings.DRIVER_PATH, provider=settings.DRIVER_PROVIDER)
+    key_names = driver.key_names
+
+    out = []
+    for dataset in datasets:
+        metadata = filter_metadata(driver.get_metadata(dataset), columns)
+        metadata["keys"] = OrderedDict(zip(key_names, dataset))
+        out.append(metadata)
```
dionhaefner, line 52: There's a `with driver.connect()` missing around this block. Without it, the connection is opened and closed for every dataset which should have a decisive performance impact.

terracotta/server/metadata.py
```diff
@@ -50,6 +50,35 @@ class Meta:
...
+            if value == "[]":
+                return []
+
+            return value[1:-1].split(", ")
+        except ValueError:
+            raise ValidationError("Invalid input for a list of values.")
+
+
+class MetadataColumnsSchema(Schema):
+    columns = CommaSeparatedListField(description="Columns of dataset to be returned")
```
dionhaefner, line 67: I think `fields.List(fields.String())` will do here, should be easy enough to jsonify that on the frontend.

DenizYil: Will do. I was unsure if you wanted the quotations or not (based on the example provided) - but that makes more sense :)

DenizYil: Actually, it doesn't seem to be that simple 😦 

All the queries are strings, so we'd have to parse this string `["metadata", "bounds"]`. I'm not really sure what would be the best way to tackle this.

We could use JSON parsing or `ast.literal_eval` - but they both seem a bit unideal. Each solution requires its own field, anyway - unless we want to add this functionality to the handler itself (which I personally don't think we should do).

I'll wait for your advice before progressing. Perhaps I'm missing something obvious.

terracotta/handlers/metadata.py
DenizYil: Great catch - didn't know - thanks 😄

terracotta/server/metadata.py
dionhaefner: I don't get it. `fields.List(fields.String())` should work just fine on requests like this:

```
POST /metadata?columns=["metadata","bounds"]
```

terracotta/handlers/metadata.py
dionhaefner: Did this solve the performance issues? No need to benchmark it in production, the test you had should suffice.

terracotta/server/metadata.py
DenizYil: `POST http://localhost:5000/metadata?columns=["metadata", "bounds"]`

```py
class MetadataColumnsSchema(Schema):
    columns = fields.List(
        fields.String(),
        description="List of columns to return",
        required=False,
    )
```
Handler:
```py
columns_schema = MetadataColumnsSchema()
columns = columns_schema.load(request.args)
```

Calling that raises the following exception:
```py
marshmallow.exceptions.ValidationError: {'columns': ['Not a valid list.']}
```
This makes sense, because: 
```py
print(request.args)
print(request.args.get("columns"))
print(type(request.args.get("columns")))
```
prints the following
```
ImmutableMultiDict([('columns', '["metadata", "bounds"]')])
["metadata", "bounds"]
<class 'str'>
```
---
We could go with using JSON to parse the string first into a list or `ast`. What would you prefer? Something along these lines:
```py
columns_str = request.args.get("columns")
try:
    columns_list = json.loads(columns_str)
except json.JSONDecodeError:
    ...
    
columns = columns_schema.load({"columns": columns_list})
```
or
```py
class JSONStringListField(fields.Field):
    def _deserialize(self, value, attr, data, **kwargs):
        try:
            return json.loads(value)
        except json.JSONDecodeError:
            raise ValidationError("Invalid JSON string format.")
```
What do you think?

dionhaefner: I see. Other routes handle this via JSON: 

https://github.com/DHI-GRAS/terracotta/blob/3b28c08852e66d9cceeb34ed6c458f934d36e2d6/terracotta/server/colormap.py#L46-L58

You can just copy this pattern.

DenizYil: Cheers! It works as expected now 😄

dionhaefner:
Starting to look good! Let's get that runtime config implemented and test coverage for the `POST /metdata` endpoint, then we should be good to go.

terracotta/handlers/metadata.py
```diff
@@ -47,7 +47,7 @@ def multiple_metadata(
 
     out = []
     with driver.connect():
-        for dataset in datasets:
+        for dataset in datasets[:settings.MAX_POST_METADATA_KEYS]:
```
dionhaefner: Can we throw a 400 error instead? Silently returning less data than requested will net us some angry issues :)

DenizYil: Good catch 😄

terracotta/server/metadata.py
```diff
@@ -72,7 +114,53 @@ def get_metadata(keys: str) -> Response:
...
+            schema: MetadataColumnsSchema
+          - in: body
+            schema: MultipleMetadataDatasetsSchema
+        responses:
+            200:
+                description: All metadata for given dataset
+                schema: MetadataSchema
+            400:
+                description:
+                    If the MAX_POST_METADATA_KEYS (100 by default) limit is exceeded
```
dionhaefner, line 148: Mentioning the default here could confuse end users (since the deployment they're looking at could have a completely different limit).

```suggestion
                    If the maximum number of requested datasets is exceeded
```

terracotta/server/metadata.py
```diff
@@ -72,7 +114,53 @@ def get_metadata(keys: str) -> Response:
...
+            404:
+                description: No dataset found for given key combination
+    """
+    from terracotta.handlers.metadata import multiple_metadata
+
+    request_body = request.json
+    if not isinstance(request_body, dict):
+        raise InvalidArgumentsError("Request body must be a JSON object")
```
dionhaefner, line 156: Please add a test that covers this case (malformed request body).

terracotta/handlers/metadata.py
```diff
@@ -3,18 +3,60 @@
...
+def multiple_metadata(
+    columns: Optional[List[str]], datasets: List[List[str]]
+) -> List[Dict[str, Any]]:
+    """Returns all metadata for multiple datasets"""
+    settings = get_settings()
+    driver = get_driver(settings.DRIVER_PATH, provider=settings.DRIVER_PROVIDER)
+    key_names = driver.key_names
+
+    if len(datasets) > settings.MAX_POST_METADATA_KEYS:
+        raise InvalidArgumentsError(
```
dionhaefner, line 50: Please add a test that covers this.

dionhaefner (changes requested):
Some last nits and we're good to merge. Thanks again @DenizYil

DenizYil:
Should be good now @dionhaefner! Thank you so much for your thorough feedback and your patience :-) I know I made some silly mistakes, but I've learned a good amount from this, so thanks a lot 😄 

This will be a very nice feature to have 🥳

dionhaefner (approved):
Looking great, thanks!

I didn't see any silly mistakes here. I think this PR was a display of healthy OSS dynamics at work. Thanks for contributing.

PR #313, opened 2023-08-14
Create docker image

charalamm:
Added code for creating a docker image to be used when deploying the terracotta server. It has been used in an azure container app and it had really good results.


charalamm:
@dionhaefner 

> What's the concrete use case for this? It looks like it's meant for production, but why use terracotta serve then instead of a real webserver with concurrency?
The intention is to use the docker image in a containerized deployment. I am going to use that image with azure container apps.
Good point about the real server. I added `gunicorn` but not sure if it is enough. Do you mean servers like nginx?

dionhaefner (changes requested):
Thanks, this looks better now. 

1. Could we please move the makefile and dockerignore under a `docker/` subfolder to keep this self-contained? (You might have to move the build context one level deeper)
2. Please add a minimal example on how to use this to `docker/README.md`.

charalamm:
> 1. Could we please move the makefile and dockerignore under a `docker/` subfolder to keep this self-contained? (You might have to move the build context one level deeper)
> 2. Please add a minimal example on how to use this to `docker/README.md`.

Thanks @dionhaefner for the recommendations, I added them.

dionhaefner:
LGTM, thanks

dionhaefner (approved):

Issue #314, opened 2023-08-24
terracotta connect requesting metadata keys in wrong order


rumski20:
When I run `terracotta connect` on an instance of terracotta server (eg. `localhost:5000`), I get a bunch of 404 errors in the console:
```
XHRGET
http://localhost:5000/metadata/cc/d30/WHNGS/2020
[HTTP/1 404 NOT FOUND 10ms]
```
http://localhost:5000/metadata/d30/cc/2020/WHNGS is the correct order as indicated by http://localhost:5000/keys

I can probably fix this locally, if you can point me to where these `/metadata` requests are being generated.

dionhaefner:
Ah, this is what I suspected to be the cause of #305 and #310. Good to see it confirmed!

@DenizYil do you think you could take a stab at this? I remember you mentioning that you felt confident to do work on the client.

DenizYil:
Yes, absolutely! I noticed it too and was planning on fixing it along with the other changes to the client 😄

dionhaefner:
@DenizYil I was planning to release terracotta 8.1 as soon as possible. Do you think we could add this fix, potentially without the other changes to the client?

dionhaefner:
Unfortunately not.

I'm contemplating whether we should rip out the react app and go back to the bare-bones vanilla Javascript client I wrote by hand. That one I can at least maintain myself. I see little value in keeping components that provide virtually no functional improvements and are not actively maintained.

DenizYil:
I'd like to really apologize for the delay with the PR. For context, I started my master's on the 1st and have severely cut the number of hours I'm working. I also completely forgot about the PR until I received a notification for this, which is completely on me. I'll look at the errors now and hopefully, we should be able to merge it today or tomorrow. 

Sincere apologies for the delay.

> Unfortunately not.
> 
> I'm contemplating whether we should rip out the react app and go back to the bare-bones vanilla Javascript client I wrote by hand. That one I can at least maintain myself. I see little value in keeping components that provide virtually no functional improvements and are not actively maintained.
This could definitely be an avenue we could consider. I haven't seen the old bare-bones client, but I'm a bit against it considering that after merging the PR, the client should be good for a long time. I would also like to put extra effort into maintaining it moving forward if that is of course okay with you.

[...]

Issue #312, opened 2023-08-09
Update terracotta interactive web client dependencies 


DenizYil:
I noticed that the web-viewer is using quite outdated dependencies and that the code could be cleaned up quite a bit. Material UI upgraded to version 5 a few years ago and has since made large improvements. It's now known as Mui.

There are some dependencies that should be moved under `devDependencies` as well, such as `typescript` and `@types/geojson`.

dionhaefner:
I wouldn't call this a priority but of course you're welcome to do this :+1:

PR #317, opened 2023-09-03
refactor: client app to use latest specifications

DenizYil:
- Updates all packages used to their latest version. This included bumping and migrating Material UI from version 4 to 5. This change in itself is quite large and is what the majority is for. It required all files to be updated. 
  - This creates a bit of a visual difference, but it follows their guidelines.
- Also updated the theme specifications to use the latest DHI ones. This includes some small changes, e.g. font sizes, margins, etc. This also included some visual changes like changing the text field variants, etc.
- Added Eslint so the code is consistent and follows the latest DHI practices.
- Fixed a bug where the keys were often not used in the correct order
- Tons of small optimizations here and there, plenty of work still left in this regard though.
- Removed unused images
- Removed custom webpack settings/compiler and used the default from CRA. As a future change, we could look into using Vite or something else. 
- Closes https://github.com/DHI-GRAS/terracotta/issues/314 and https://github.com/DHI-GRAS/terracotta/issues/312
- .. and more

@dionhaefner I had some issues running the connect command. If you're able to try that command (when giving the client a look), that would be really appreciated - but it should all still work.


dionhaefner:
Thanks @DenizYil :tada:

I'm seeing some typescript errors when [trying to build the app on CI](https://github.com/DHI-GRAS/terracotta/actions/runs/6071290946/job/16469042114?pr=317). Could you have a look at that?

DenizYil:
@dionhaefner sincere apologies for the delay with this. I had somehow completely missed all of this.

All the ESLint fixes should be fixed now and the build seems to complete successfully. Please let me know if there's anything else needed for this. I'll monitor this PR actively.

I'm not exactly sure why the Python pipelines are failing, but I guess that is more your expertise than mine (please let me know if there's any way that I can help).

dionhaefner:
Unfortunately there seem to be more problems with the client. The basemap is broken, and the proportions look off (font size too large and margins too small).

### Before

Screenshot 2023-09-20 at 14 02 20

### After

Screenshot 2023-09-20 at 14 02 34

---

Is there any way you could implement only the bug fix to #314 in a separate PR that we can merge right away, then keep iterating on this?

DenizYil:
> Is there any way you could implement only the bug fix to #314 in a separate PR that we can merge right away, then keep iterating on this?

Done in https://github.com/DHI-GRAS/terracotta/pull/318. Good call.

--- 

For the feedback: I'll fix the basemap, not sure how that disappeared as it works locally, but will investigate tomorrow. For the proportions, it uses the latest DHI (and MUI) specifications, so some numbers were increased (e.g. base fonts). I wasn't sure whether or not to try to match the old one 1:1, but if that's what we decide, I can hardcode the numbers and change the styling, which should be straightforward and a relatively quick thing to change throughout.

dionhaefner:
I'm not married to the original font size, but if we end up changing it we should also adjust some margins here and there to give the design some breathing room. Right now the UI looks crammed to me, especially around the contrast slider and in / around the "raster URL" / "Metadata" boxes.

PR #318, opened 2023-09-21
fix(client): key order

DenizYil:
Closes https://github.com/DHI-GRAS/terracotta/issues/314 as discussed in https://github.com/DHI-GRAS/terracotta/pull/317.


dionhaefner:
Awesome, thanks!

PR #322, opened 2023-12-07
Add rgb_range percentile support

atanas-balevsky:
Add support for using percentiles instead of absolute values for the `/rgb`'s endpoint `*_range` parameters.

gives us the possibility to request the `/rgb?r_range=p2,p97...` instead of calling `/metadata` first to fetch the pct's absolute values in order to construct an `/rgb` url

### DIFF
--- a/terracotta/handlers/rgb.py
+++ b/terracotta/handlers/rgb.py
@@ -11,7 +11,8 @@ from terracotta import get_settings, get_driver, image, xyz, exceptions
 from terracotta.profile import trace
 
 Number = TypeVar("Number", int, float)
-ListOfRanges = Sequence[Optional[Tuple[Optional[Number], Optional[Number]]]]
+NumberOrString = TypeVar("NumberOrString", int, float, str)
+ListOfRanges = Sequence[Optional[Tuple[Optional[NumberOrString], Optional[NumberOrString]]]]
 
 
 @trace("rgb_handler")
@@ -90,10 +91,10 @@ def rgb(
             scale_min, scale_max = band_stretch_override
 
             if scale_min is not None:
-                band_stretch_range[0] = scale_min
+                band_stretch_range[0] = get_scale(scale_min, metadata)
 
             if scale_max is not None:
-                band_stretch_range[1] = scale_max
+                band_stretch_range[1] = get_scale(scale_max, metadata)
 
             if band_stretch_range[1] < band_stretch_range[0]:
                 raise exceptions.InvalidArgumentsError(
@@ -105,3 +106,20 @@ def rgb(
 
     out = np.ma.stack(out_arrays, axis=-1)
     return image.array_to_png(out)
+
+
+def get_scale(scale: NumberOrString, metadata) -> Number:
+    if isinstance(scale, (int, float)):
+        return scale
+    if isinstance(scale, str):
+        # can be a percentile
+        if scale.startswith("p"):
+            # TODO check if percentile is in range
+            percentile = int(scale[1:]) - 1
+            return metadata["percentiles"][percentile]
+
+        # can be a number
+        return float(scale)
+    raise exceptions.InvalidArgumentsError(
+        "Invalid scale value: %s" % scale
+    )
--- a/terracotta/handlers/rgb.py
+++ b/terracotta/handlers/rgb.py
@@ -91,11 +91,12 @@ def rgb(
             band_stretch_range = list(metadata["range"])
             scale_min, scale_max = band_stretch_override
 
+            percentiles = metadata.get("percentiles", [])
             if scale_min is not None:
-                band_stretch_range[0] = image.get_stretch_scale(scale_min, metadata)
+                band_stretch_range[0] = image.get_stretch_scale(scale_min, percentiles)
 
             if scale_max is not None:
-                band_stretch_range[1] = image.get_stretch_scale(scale_max, metadata)
+                band_stretch_range[1] = image.get_stretch_scale(scale_max, percentiles)
 
             if band_stretch_range[1] < band_stretch_range[0]:
                 raise exceptions.InvalidArgumentsError(
--- a/terracotta/handlers/rgb.py
+++ b/terracotta/handlers/rgb.py
@@ -10,9 +10,10 @@ from concurrent.futures import Future
 from terracotta import get_settings, get_driver, image, xyz, exceptions
 from terracotta.profile import trace
 
-Number = TypeVar("Number", int, float)
 NumberOrString = TypeVar("NumberOrString", int, float, str)
-ListOfRanges = Sequence[Optional[Tuple[Optional[NumberOrString], Optional[NumberOrString]]]]
+ListOfRanges = Sequence[
+    Optional[Tuple[Optional[image.NumberOrString], Optional[image.NumberOrString]]]
+]
 
 
 @trace("rgb_handler")
@@ -91,10 +92,10 @@ def rgb(
             scale_min, scale_max = band_stretch_override
 
             if scale_min is not None:
-                band_stretch_range[0] = get_scale(scale_min, metadata)
+                band_stretch_range[0] = image.get_stretch_scale(scale_min, metadata)
 
             if scale_max is not None:
-                band_stretch_range[1] = get_scale(scale_max, metadata)
+                band_stretch_range[1] = image.get_stretch_scale(scale_max, metadata)
 
             if band_stretch_range[1] < band_stretch_range[0]:
                 raise exceptions.InvalidArgumentsError(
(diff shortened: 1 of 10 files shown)


terracotta/server/rgb.py
```diff
@@ -29,25 +29,34 @@ class Meta:
     g = fields.String(required=True, description="Key value for green band")
     b = fields.String(required=True, description="Key value for blue band")
     r_range = fields.List(
-        fields.Number(allow_none=True),
+        fields.String(allow_none=True, validate=validate.Regexp("^p?(\d*\.)?\d+$")),
```
dionhaefner, line 32: This is not the correct format.

Previously, this accepted any number, which includes scientific notation like `1.3E-2`. Also, percentiles must be ints between 0 and 100 and cannot be float numbers.

To prevent regressions I would advocate for still using `fields.Number` under the hood, with some additional logic to detect the percentile case. A [custom field](https://marshmallow.readthedocs.io/en/stable/custom_fields.html) class should do, similar to [what is described here](https://stackoverflow.com/questions/61614546/python-marshmallow-field-can-be-two-different-types).

tests/handlers/test_rgb.py
```diff
@@ -74,7 +74,10 @@ def test_rgb_lowzoom(use_testdb, raster_file, raster_file_xyz_lowzoom):
 
 
 @pytest.mark.parametrize(
-    "stretch_range", [[0, 20000], [10000, 20000], [-50000, 50000], [100, 100]]
+    "stretch_range", [
+        [0, 20000], [10000, 20000], [-50000, 50000], [100, 100],
+        ["0", "20000"], ["10000", "20000"], ["-50000", "50000"], ["100", "100"],
```
dionhaefner, line 79: What does this do?

dionhaefner (commented):
Thanks for the PR.

There are some issues with it that look fixable. More importantly, I want at least one additional contributor to weigh in whether this is a good idea or not.

The interface is getting pretty overloaded, as evidenced by the type annotations:

```python
ListOfRanges = Sequence[Optional[Tuple[Optional[NumberOrString], Optional[NumberOrString]]]]
```

Does anyone else have mixed feelings about this?

atanas-balevsky:
Dear @dionhaefner ,
Thanks for the feedback! i've addressed some aspects of it.

now it's much clear that there can be a string or number. The regex is gone and there's more natural way of inferring the range type. i also added a test for the scientific notation when doing absolute values (didn't know you're supporting these)
i'm not sure why the tests action is not running now, but i believe it should be green

mrpgraae:
> Thanks for the PR.
> 
> There are some issues with it that look fixable. More importantly, I want at least one additional contributor to weigh in whether this is a good idea or not.
> 
> The interface is getting pretty overloaded, as evidenced by the type annotations:
> 
> ```python
> ListOfRanges = Sequence[Optional[Tuple[Optional[NumberOrString], Optional[NumberOrString]]]]
> ```
> 
> Does anyone else have mixed feelings about this?

I may be biased since @atanas-balevsky is a colleague of mine, and we need this feature for something we are building over at SkyFi. Still, I think this is a good example of leveraging the fact that we have pre-computed information about an image to provide "magic" features like this, so I'd like to convince you that I think it's a good idea, even without my bias 🙂

Regarding the type, I think it was already quite complicated before. The only difference in this PR is that `Number` changed to `NumberOrString`. But if you think it would be better to do the percentiles as a separate query parameter, to not overload this parameter so much, then I don't see anything wrong with that. It would also help make the type validation simpler.

terracotta/server/rgb.py
```diff
@@ -21,6 +22,36 @@ class RGBQuerySchema(Schema):
...
 
+def validate_range(data):
+    if isinstance(data, str) and data.startswith("p"):
+        if not re.match("^p\d+$", data):
+            raise ValidationError("Percentile format is `p<digits>`")
+    else:
+        try:
+            float(data)
+        except ValueError:
+            raise ValidationError("Must be a number")
```
dionhaefner, line 33: Can we use the same exception for both branches that contains the expected format?

terracotta/handlers/rgb.py
```diff
@@ -105,3 +106,20 @@ def get_band_future(band_key: str) -> Future:
...
+        # can be a percentile
+        if scale.startswith("p"):
+            # TODO check if percentile is in range
+            percentile = int(scale[1:]) - 1
+            return metadata["percentiles"][percentile]
+
+        # can be a number
+        return float(scale)
+    raise exceptions.InvalidArgumentsError(
+        "Invalid scale value: %s" % scale
```
dionhaefner, line 124: Use f-string.

terracotta/handlers/rgb.py
```diff
@@ -105,3 +106,20 @@ def get_band_future(band_key: str) -> Future:
...
+        return scale
+    if isinstance(scale, str):
+        # can be a percentile
+        if scale.startswith("p"):
+            # TODO check if percentile is in range
+            percentile = int(scale[1:]) - 1
+            return metadata["percentiles"][percentile]
+
+        # can be a number
+        return float(scale)
```
dionhaefner, line 122: Can it though? I don't think passing a string should be valid if not using the percentile feature.

dionhaefner (changes requested):
Alright, I'm willing to merge this if we can address the issues above, plus:

1. Add full type annotations to all functions + arguments.
2. Ensure pre-commit hooks pass.
3. Add the same argument to `/singleband`.
4. Push coverage to the target.

Thanks @atanas-balevsky.

terracotta/server/rgb.py
```diff
@@ -21,6 +22,36 @@ class RGBQuerySchema(Schema):
     tile_x = fields.Int(required=True, description="x coordinate")
 
 
+def validate_range(data):
+    if isinstance(data, str) and data.startswith("p"):
+        if not re.match("^p\d+$", data):
+            raise ValidationError("Percentile format is `p<digits>`")
```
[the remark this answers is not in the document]
atanas-balevsky: Hey @dionhaefner here we'd do only validation of the format. if the number is in range we'd see when we try to access the metadata percentiles: https://github.com/DHI-GRAS/terracotta/pull/322/files#diff-f3ee2be066aa9a6d1432c4e13259b840360f0bf46320bf3b5d8cc07015cf9d09R198-R203 (may save in some cases where metadata[percentiles] is not 0..100

atanas-balevsky:
> Alright, I'm willing to merge this if we can address the issues above, plus:
> 
> 1. Add full type annotations to all functions + arguments.
> 2. Ensure pre-commit hooks pass.
> 3. Add the same argument to `/singleband`.
> 4. Push coverage to the target.
> 
> Thanks @atanas-balevsky.

Dear @dionhaefner, 
thanks for the feedback, i believe i addressed all the points. can we rerun the tests & reports?

Happy holidays!
Nas

terracotta/server/fields.py
```diff
@@ -0,0 +1,37 @@
...
+Custom marshmallow fields for the server API.
+"""
+
+import re
+from marshmallow import ValidationError, fields
+
+from typing import Any, Union
+
+
+class StringOrNumber(fields.Field):
```
dionhaefner, line 12: Please add a docstring

terracotta/server/fields.py
```diff
@@ -0,0 +1,37 @@
...
+    ) -> Union[str, float, None]:
+        if isinstance(value, (str, bytes)):
+            return fields.String()._deserialize(value, attr, data, **kwargs)
+        elif isinstance(value, (int, float)):
+            return fields.Float()._deserialize(value, attr, data, **kwargs)
+        else:
+            raise ValidationError("Must be a string or a number")
+
+
+def validate_stretch_range(data: Any) -> None:
```
dionhaefner, line 34: Please add a docstring

terracotta/server/rgb.py
```diff
@@ -29,25 +30,43 @@ class Meta:
...
-        fields.Number(allow_none=True),
+        StringOrNumber(allow_none=True, validate=validate_stretch_range),
         validate=validate.Length(equal=2),
         example="[0,1]",
         missing=None,
-        description="Stretch range [min, max] to use for red band as JSON array",
+        description=(
+            "Stretch range [min, max] to use for the red band as JSON array. "
+            "Min and max may be numbers to use as absolute range, or strings "
+            "of the format p<int> with an integer between 0 and 100 to use "
```
dionhaefner, line 40: Don't use `<>`, angled brackets get parsed as HTML in the API docs. Either escape with `&lt;` and `&gt;` or use a different character.

dionhaefner (changes requested):
Thanks, this looks really good! Some final nits.

tests/server/test_fields.py
```diff
@@ -0,0 +1,42 @@
+from marshmallow import ValidationError
+import pytest
+from terracotta.server.singleband import SinglebandOptionSchema
+
+
+@pytest.mark.parametrize(
+    "args, expected",
+    [("[0, 1]", [0, 1]), ('["p2", "p28"]', ["p2", "p28"]), (None, None)],
+)
+def test_serde(args, expected):
```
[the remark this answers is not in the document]
atanas-balevsky: serialization/deserialization, had to test the deserialization part of the custom marshmallow types to fit the codecov

atanas-balevsky:
Dear @dionhaefner!
Thanks for the feedback. i believe i addressed most of the comments. can we rerun the test actions?

dionhaefner (approved):
LGTM, thanks!

PR #324, opened 2023-12-14
Remove numpy version upper bound restriction

serj90:
We don't need numpy<1.24 restriction as crick==0.0.4 is released https://pypi.org/project/crick/0.0.4/


dionhaefner:
Thanks!

PR #331, opened 2024-02-27
Update docker image to python 3.12

charalamm:


charalamm:
@dionhaefner yes good idea

charalamm:
@dionhaefner Unfortunately we dont  have dockerhub account and it seems it will take some time to create one. I will add the CI once we have it. I am de-drafting so you can merge if it looks good to you too

dionhaefner:
Oh I forgot that DockerHub isn't free anymore. Let's merge this first, but testing the image build on CI could be valuable too, to ensure nothing breaks. I'll add an issue for that.

PR #336, opened 2024-03-27
Build push a docker image to dockerhub on release

charalamm:


dionhaefner:
Awesome. Did you test this workflow somewhere @charalamm? (No is fine, I'd just like to know)

charalamm:
@dionhaefner I tested it only when pushing on a branch and without the commit tag. Now that I am thinking about it I could test it in a throwaway repo. I will do that and then we can merge I guess.

dionhaefner:
We could also merge this and do a post-release to test the deployment. Not quite as elegant but if it saves you work I'd be up for it.

charalamm:
Hello @dionhaefner I have tested it in separate repo and it works and uses the release tag as the docker image tag. It also updates the latest tag image. I am merging if that is ok

Tests are failing for irrelevant reasons as far as I can see

dionhaefner:
Sounds great, thanks @charalamm !

PR #338, opened 2024-06-03
Do not overwrite GDAL_CACHEMAX if set

### LINKED ISSUE #337: Configure GDAL_CACHEMAX when running optimize-rasters
I've been tweaking this line and bumping `CACHEMAX` to several GB seems to give a nice performance improvement when I have just a few large rasters to process and RAM to spare:

https://github.com/DHI/terracotta/blob/018f96341bbfcd3876f6f89a30b0fc59ec855599/terracotta/scripts/optimize_rasters.py#L35

Is there a preferred way to make this configurable? Could be a check for existing GDAL_CACHEMAX environment variable and fall back to 512MB as a safe enough default?

dionhaefner:
Fixes #337

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -144,7 +144,7 @@ jobs:
           key:
             ${{ runner.os }}-conda-${{ matrix.python-version }}-${{ hashFiles('environment.yml') }}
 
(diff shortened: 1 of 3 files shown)


j08lue:
(time to update https://github.com/conda-incubator/setup-miniconda to v3? 🤷 )

PR #345, opened 2024-09-18
Added time based filtering to ingest and optimize_rasters functions

mmarfat:
### Description
This PR introduces the `--ignore-older-than` flag to both the `optimize_rasters()` and `ingest()` functions, allowing users to skip processing files that are older than a specified cutoff based on their last modified time.

### Key Changes
- **New `--ignore-older-than` flag:**  
  Enables users to set a time threshold, ignoring files older than the provided value.

- **Supported formats for the `--ignore-older-than` flag:**
  - Time durations: `5d`, `2h`, `30m`, `10s` (representing days, hours, minutes and seconds, respectively)
  - Specific timestamps: `YYYY-MM-DD HH:MM:SS`

### Behavior
- Files with a last modified time older than the given threshold will be skipped and not optimized or ingested.

### DIFF
--- a/terracotta/scripts/click_types.py
+++ b/terracotta/scripts/click_types.py
@@ -2,9 +2,10 @@
 
 Custom click parameter types and utilities.
 """
-
 from typing import List, Any, Tuple, Dict
+from datetime import datetime, timedelta
 import pathlib
+import contextlib
 import glob
 import re
 import os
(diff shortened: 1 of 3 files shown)


dionhaefner:
Thanks @mmarfat but I'm not sure I get it. Why does Terracotta need this feature?

mmarfat:
I suppose Terracotta itself doesn't really "need" this feature, rather it's something that was useful for my specific application. As far as I can tell there is no explicit way to define which files should be ingested (other than trough the naming pattern).

In my specific case this was limiting because I'm working with a folder over which I have no control and it's gotten quite big over time. This causes the ingest operation to be quite slow. To avoid ingesting all the files all the time I implemented this so it would only ingest the "newest" files.

I know it's not the most attractive feature but since it's not taking away any functionality I thought I'd suggest it for a PR.

dionhaefner:
Hmm I see. Any reason why you didn't want to reach for the [Python API](https://terracotta-python.readthedocs.io/en/latest/get-started.html#using-the-python-api) for this?

I'm not strongly opposed to this feature and the implementation looks high quality, but I'm trying to understand if there's a genuine blind spot that's addressed here or whether it adds complexity for little net-gain.

j08lue:
I was thinking the same - even considering the "batteries included" philosophy of Terracotta, this looks like a case for writing your own little CLI script that implements your specific file filtering logic.

Also agree that this looks very nicely made!

So I'd recommend not to implement this, since

1. The logic is quite specific to your use case
2. It would add complexity to users looking at the CLI to understand and decide whether they want to use this - and in most cases probably not
3. Adds a lot of code (+tests)
4. There is another simple way to achieve this (custom script)

But keep it as a very nice reference how to solve this that hopefully others will find useful - and if we get more reactions, we can reconsider.

my 2ct

mmarfat:
Fair enough, I agree that this could be considered too specific to my use case and therefore not really in style of the current lightweight implementation. If someone finds this useful, they can always refer to my [fork](https://github.com/mmarfat/terracotta/).

Regardless, thanks for the discussion @j08lue & @dionhaefner

PR #346, opened 2024-10-22
make terracotta python 3.12 compatible

jkittner:
fix deprecation/removal of pkg_resources and rasterio is_tiled

if we were to drop 3.8 which is now EOL we wouldn't need the additional dependency of `importlib_resources` and the if/else imports.

PR #347, opened 2024-10-22
remove upper bound from numpy version

jkittner:
https://github.com/dask/crick/issues/53 was resolved and tests pass locally for me.

This is causing issues for us since other packages started requiring `numpy>=2.0`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -79,6 +79,7 @@ setup(
         "tqdm",
     ],
     extras_require={
+        ':python_version == "3.9"': ["numpy<2.0.0"],
         "test": [
             "pytest",
             "pytest-cov",
--- a/setup.py
+++ b/setup.py
@@ -18,9 +18,6 @@ long_description = re.sub(
 numpy_constraints = (
     ">=1.15",
     "!=1.17.0",
-    # TODO: drop this once all deps support numpy 2.0
-    # see https://github.com/dask/crick/issues/53
-    "<2",
 )
 numpy_version = ",".join(numpy_constraints)
 


jkittner:
#346 should resolve those 3.11 test failures - it looks like it captures the rasterio warnings?

j08lue:
We can check once that one is merged and we rebase here.

jkittner:
There are no 3.9 wheels for crick on pypi. It tries to install version 0.0.6.

```
2024-10-24T10:32:57.6608536Z   Downloading crick-0.0.6-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (1.1 kB)
```
However, version 0.0.7 is available - but only with wheels starting at 3.10 since crick 0.0.7 dropped support for 3.9. https://github.com/dask/crick/blob/5b7f0c779ac873dfc895f39d41a6682e519764b6/pyproject.toml#L17 (see: https://github.com/dask/crick/pull/57)

The only way to do this would be this:

```diff
diff --git a/setup.py b/setup.py
index 5b16c7e..be42428 100644
--- a/setup.py
+++ b/setup.py
@@ -79,6 +79,7 @@ setup(
         "tqdm",
     ],
     extras_require={
+        ':python_version == "3.9"': ['numpy<2.0.0'],
         "test": [
             "pytest",
             "pytest-cov",
```

which would mean terracotta would use numpy < 2 and crick 0.0.6 on 3.9 and numpy > 2 and crick > 0.0.6 on > 3.9 - I think that would be fine.

Tbh, quite an annoying mess these dependencies - sorry...

j08lue:
y tho
- https://github.com/dask/crick/pull/58

jkittner:
macos is missing the postgres binaries: `pg_config` (or they are not on `PATH`) that's why it fails compiling `psycopg2` and building a wheel.
At least widnows and 3.9 works now...

PR #348, opened 2024-10-31
add turbo colormap

jkittner:
This PR adds the popularly used "turbo" color map as a builtin/natively supported colormap. Turbo should be used over "jet". There have been somewhat recent studies highlighting the advantage: https://research.google/blog/turbo-an-improved-rainbow-colormap-for-visualization/ .

I hope I did everything needed - is the static docs file generated automagically when the docs are deployed or any manual work needed?

I generated the files using the `generate_cmaps.py` script. It yielded a few more additions and a few changes - not sure we want to update/add those? Let me know!

see also:
- https://github.com/matplotlib/matplotlib/issues/15091/
- https://github.com/matplotlib/matplotlib/pull/15275


dionhaefner:
Yes, please re-run `generate_cmaps.py` and commit all changes.

You don't need to pitch the advantages of turbo or any specific color map for that matter, it's just that we're slightly out of sync with matplotlib since that script has to be run manually :)

jkittner:
Okay, I reran the script and added all changes. I had a look at why the client app build is failing - I feel like that does not work for a pull request coming from a fork? It either can't check out my branch or when not specifying a `ref` it can't push back the changes - not sure If we need to change that?

dionhaefner:
Can you try removing this from the `build-client.yml` action?

```yaml
with:
    ref: ${{ github.head_ref }}
```

dionhaefner:
I see, thanks. Should be fine.

Unrelated: I really dislike this enumeration of colormaps in the client. After all, we have the `/colormaps` to retrieve this information directly from the server. Alas, a problem for another day.

dionhaefner (approved):

Issue #23, opened 2018-07-10
Support non-linear rescaling


dionhaefner:
Such as conversions between different color spaces

PR #350, opened 2024-11-08
Optional gamma correction parameter

biserhong:
This PR adds gamma correction functionality to the `/singleband` and `/rgb` endpoints. It is requested by adding an optional `gamma_factor` parameter. 

An important observation about enabling this functionality is that the exponential function preserves the relative percentile ranges in the [0, 1] interval. So the computed percentiles in the metadata can be carried through after gamma correcting a tile.


dionhaefner:
Thanks for the implementation. Can you say a bit more about the use case for this? Any reason why this can't be done offline (during preprocessing) or using `/compute`?

biserhong:
> Thanks for the implementation. Can you say a bit more about the use case for this? Any reason why this can't be done offline (during preprocessing) or using `/compute`?

We use terracotta to display images customers purchase at SkyFi. For true colour 3-band composites, rendering the data linearly does not provide a representation that is similar to what the human eye would perceive. Applying a gamma correction between 1.6 and 2.2, on an otherwise linear colour space makes for something that is a lot closer to a “natural colour composite”. We also want the preview on the map to be as close as possible to the png previews we send customers along with the raster, to which gamma correction is applied.

Regarding the option of doing it offline, we would like to offer users the ability to adjust min/max and gamma factor on the fly. The `/compute` endpoint also only outputs a single band image and cannot use the precomputed metadata for color stretching. 

I also think gamma correction would be a good addition to the capabilities of terracotta since it’s a standard and common processing technique, and its implementation doesn’t really affect any other functionality.

vincentsarago:
FYI in rio-tiler/titiler we use https://github.com/vincentsarago/color-operations (fork of mapbox/rio-color), Just sharing if you want to support more `color` operations in the future ;-) 

https://github.com/cogeotiff/rio-tiler/blob/dd5ce03f8f463ec93f3a8787d7155d318397cd7c/rio_tiler/models.py#L524-L536

dionhaefner:
Thanks @vincentsarago! I like this a lot, and a feature like this has been on the roadmap for a long time: #23.

I'd strongly prefer an implementation that doesn't assume a specific color transformation like gamma, and instead implements support for a range of transforms (like the ones provided via `color-operations` / `rio-color`). @biserhong do you think that would be feasible?

biserhong:
@vincentsarago @dionhaefner To be honest, I also used `rio-color`'s implementation of the same functionality but had to change it a bit to handle the `float` case (so `color-operations` actually has almost identical functions). 

However, I added `color-operations` as a dependency to use your implementation. 

Some notes:
- You still need a wrapper `gamma_correction` function to handle normalization and rescaling before stretching. 
- As far as I can tell we essentially cannot use the operations parsing because `terracotta` is predicated on directly working only on single bands.
- To be able to string together several color operations and still reuse the precomputed percentiles would probably have to be thought through as well to see if it would work in all cases.

dionhaefner:
Thanks @biserhong !

I'm hesitant to introduce features that complicate the API for a singular use case like gamma correction. I can see how this is probably fine for the few most important use cases, but inevitably leads to a kitchen sink of parameters over time that becomes a nightmare to clean up / refactor later on.

A for me much more sustainable design would be to introduce a parameter like `/rgb/x/y/z.png?color_transform="gamma R 0.5"` that exposes a collection of methods behind a single API. Would that fit your needs?

biserhong:
> Thanks @biserhong !
> 
> I'm hesitant to introduce features that complicate the API for a singular use case like gamma correction. I can see how this is probably fine for the few most important use cases, but inevitably leads to a kitchen sink of parameters over time that becomes a nightmare to clean up / refactor later on.
> 
> A for me much more sustainable design would be to introduce a parameter like `/rgb/x/y/z.png?color_transform="gamma R 0.5"` that exposes a collection of methods behind a single API. Would that fit your needs?

It would work for me.

I'll have to check what each of the color transforms does exactly and it might turn out that for some transformation we might not be able to use the precomputed percentiles (it might also be ok, I just don't know). Are you ok with that? @vincentsarago do you have any input on this?

dionhaefner:
Percentiles are stable under any monotonous map, so we should be fine for all operations that are applied band-wise. In other cases, I think it's okay to be slightly off – the practical impact of that should be small (of course it should be documented explicitly that the given percentiles relate to untransformed data).

biserhong:
Sorry this took so long, I had to focus on some other work for a bit.

@dionhaefner Can you take a look at the latest commit and let me know if you agree with the approach? 

I'll clean up the rest of the PR and tidy up the code and variables if we agree to continue this way.

terracotta/image.py
```diff
@@ -187,6 +188,21 @@ def gamma_correction(
...
+        color_transform: str,
+        out_dtype: type = np.uint16,
+) -> Array:
+    """Apply gamma correction to the input array and scale it to the output dtype."""
+    arr = to_math_type(masked_data)
+
+    for func in parse_operations(color_transform):
+        arr = func(arr)
+
+    arr = scale_dtype(arr, out_dtype)
```
dionhaefner, line 202: This should be unnecessary no? Because we scale it to `uint8` right after?

terracotta/image.py
```diff
@@ -187,6 +188,21 @@ def gamma_correction(
...
+def apply_color_transform(
+        masked_data: Array,
+        color_transform: str,
+        out_dtype: type = np.uint16,
+) -> Array:
+    """Apply gamma correction to the input array and scale it to the output dtype."""
+    arr = to_math_type(masked_data)
+
+    for func in parse_operations(color_transform):
+        arr = func(arr)
```
dionhaefner, line 200: Does this assume a particular scaling already?

dionhaefner:
Awesome @biserhong! I like this a lot, especially how we specify stretch ranges in untransformed space which should work even with multi-band transforms.

I reckon there will be quite a bit of testing required to make sure scaling works as expected but from a UX perspective I really like it, so should be good to port to other endpoints as well.

terracotta/image.py
biserhong: Yes, you're right, the contrast stretch in `to_uint8` makes it unnecessary.

terracotta/image.py
biserhong: Yes, inputs have to be in the [0,1] range: https://github.com/vincentsarago/color-operations/?tab=readme-ov-file#color_operationsoperations

biserhong:
Thanks for the quick feedback @dionhaefner! I'll try to get something more complete this week.

terracotta/image.py
dionhaefner: I see. It may be easier to apply stretching before color transform then. I.e., normalize to `[0,1]` using `stretch_range` and clamp spillover to `0` / `1`, then apply color transform, then convert to `uint8`. That way we don't rely on a transformation of the stretch range?

biserhong: I think it may not be a good idea since it would change the result. 

The color stretch fills the [0, 1] range whereas the normalization `to_math_type` does is relative to the dtype range. So when you apply the color transform I think that would shift where values fall along the curve. 

This is what chatgpt said in the case of gamma correction:
```
Gamma correction is typically applied before color stretching, particularly when working with image data in scientific imaging, photography, or graphic design.

Here's why:

    Gamma Correction First: Gamma correction adjusts the image data to a linear color space, which compensates for the nonlinear response of human vision and many display systems. This linearization step ensures that subsequent adjustments, like color stretching, are applied to data that more accurately represents real-world light intensities.
```

dionhaefner: Sure, but who says that the dtype range is appropriate? The user-provided stretch range is telling us how to map the values in the raster to a linear scale.

Summoning @vincentsarago in case you want to weigh in on the appropriate order of linear scaling (and clamping out of range values) vs. color correction :)

biserhong:
@dionhaefner I removed all references to the gamma factor and replaced it with the color transform.

I excluded the saturation option for now because it's only valid for rgb and it's not monotonic. 

What do you think? I can proceed to tests if you agree with this approach.

terracotta/handlers/rgb.py
```diff
@@ -88,7 +90,8 @@ def get_band_future(band_key: str) -> Future:
             keys = (*some_keys, band_key)
             metadata = driver.get_metadata(keys)
 
-            band_stretch_range = list(metadata["range"])
+            band_range = list(metadata["range"])
```
dionhaefner, line 93: Why introduce this variable? Looks unused.

terracotta/server/singleband.py
```diff
@@ -65,6 +65,14 @@ class Meta:
         "hex strings.",
     )
 
+    color_transform = fields.String(
+        validate=validate_color_transform,
+        missing=None,
+        example="gamma 1 1.5, sigmoidal 1 15 0.5",
+        description="Color transform DSL string from color-operations."
+        "All color operations for singleband should specify band 1.",
```
dionhaefner, line 73: Not sure what this means. Could you elaborate?

terracotta/handlers/rgb.py
dionhaefner: Oh now I see, you're using this further below. But beware, since you're setting this in a loop only the value for the last band will be used in the color transform!

biserhong: Thanks, I had overlooked that. 

I switched things around a bit so now normalization is done outside the color transform function.

terracotta/server/singleband.py
biserhong: The DSL that the color-operations library implements has to include the band you're applying the operation to. 

As in the examples on the github page `gamma G 0.5` means apply gamma with 0.5 to the _second_ band. You can also say `gamma 2 0.5`. 

So to apply gamma or sigmoidal with the singleband route you need to specify `gamma R 0.5` or `gamma 1 0.5`. I just thought that saying `R` might be confusing in the singleband case.

biserhong:
I'll fix the tests.

biserhong:
Ok, after some back and forth, I switched things back to applying stretching first and then doing the color transform. This way the full set of color transforms is available.

All the tests are passing locally.

terracotta/server/singleband.py
dionhaefner: got it, thanks.

terracotta/handlers/rgb.py
```diff
@@ -88,7 +89,8 @@ def get_band_future(band_key: str) -> Future:
...
```
dionhaefner, line 92: Unused variable

terracotta/handlers/singleband.py
```diff
@@ -61,7 +64,8 @@ def singleband(
         out = image.label(tile_data, labels)
     else:
         # determine stretch range from metadata and arguments
-        stretch_range_ = list(metadata["range"])
+        band_range = list(metadata["range"])
```
dionhaefner, line 67: same

dionhaefner:
Thanks, this looks great now!

Could you post some screenshots please of this feature in action? As in, how does a raster look with different color transform strings via Terracotta? (I'm planning to add a section to the docs where those would come in handy.)

terracotta/handlers/singleband.py
biserhong: Fixed also.

biserhong:
darley-reef-gamma-0 6
darley-reef-gamma-1 8
darley-reef-gamma-1
darley-reef-saturation-0 5
darley-reef-saturation-1 2

@dionhaefner Are these ok? I can add more.

dionhaefner:
This looks fantastic, thank you!

Will do some final touches on docs / tests, then get this merged.

biserhong:
I update the cache action version as well.

biserhong:
One thing to note with the use of `color-operations` is that the DSL parser doesn't produce any intermediate result with operations and parameters. I couldn't figure out how to check if the operations produced are valid and you can certainly produce a string that will only fail when you try to use it.

terracotta/server/fields.py
```diff
@@ -45,3 +47,16 @@ def validate_stretch_range(data: Any) -> None:
...
+
+def validate_color_transform(data: Any) -> None:
+    """
+    Validate that the color transform is a string and can be parsed by `color_operations`.
+    """
+    if not isinstance(data, str):
+        raise ValidationError("Color transform needs to be a string")
+
+    try:
+        parse_operations(data)
```
dionhaefner, line 60: If that's the case, should we try and use the color transform here on a trivial array (such as `np.ones((1, 1))`)?

dionhaefner:
Actually @biserhong want to add some basic tests? At least in `test_image.py` and `handlers/test_singleband.py` and `handlers/test_rgb.py`. You don't need to check the results of the transforms, just to make sure everything runs as expected and produces the right shapes / dtypes.

terracotta/server/fields.py
biserhong: We could do that, it would get slightly ugly because the success of applying the transform will depend on whether it's applied on a singleband or rgb image. And `validate_color_transform` is only called from marshmallow.

I can use partial to suggest to `validate_color_transform` what the array it should test against.

dionhaefner: Ok maybe I misunderstood. What is being validated here? In which case can a transformation string pass this validation and fail later during application?

biserhong:
> Actually @biserhong want to add some basic tests? At least in `test_image.py` and `handlers/test_singleband.py` and `handlers/test_rgb.py`. You don't need to check the results of the transforms, just to make sure everything runs as expected and produces the right shapes / dtypes.

Yes, I'll add some tests as well.

terracotta/server/fields.py
biserhong: So to answer your question more fully, it does provide some validation but the parser has some shortcomings. Apart from the example I gave it will also let you pass negative numbers.

dionhaefner: Ok, so that's why I suggested we actually try and apply the operation to some fake data here (without checking the result). If someone passed an invalid parameter it will raise an exception, and if not the code will continue.

biserhong: Yes, I was mainly talking about the implementation of that idea. 

I think it works nicely now, I pushed another update.

dionhaefner:
This looks perfect, thank you very much.

PR #351, opened 2024-11-20
Install packages in docker container to enable flask profiling

adkinsjd:
Currently setting `FLASK_PROFILE=true` when using the docker container fails due to missing packages.

This PR adds the packages necessary to run flask profiling.

They are unfortunately quite dated - these are the maximum versions that I found would succeed with no other code changes. I'm happy to modify if there is a clear path profiling with more up-to-date packages.


dionhaefner (approved):
LGTM, thanks!

(I guess once more people start using those Docker images we should have them tested on CI but for now this should do.)

PR #354, opened 2025-02-17
Allows RasterIO to read GeoTiffs from S3 object stores requiring authentication

MatthewTe:
Added support for reading Geotiff files from an S3 object store that requires authentication via a secret key and access key.

Added the config variables required to enable auth S3 reading and based on the presence of these config params creates a boto3 AWSSession object to be passed through to the rasterio.Env context.

### DIFF
--- a/terracotta/drivers/geotiff_raster_store.py
+++ b/terracotta/drivers/geotiff_raster_store.py
@@ -140,10 +140,11 @@ class GeoTiffRasterStore(RasterStore):
             rio_env_options=self._RIO_ENV_OPTIONS
         )
         
-        if hasattr(settings, "RASTER_AWS_S3_ENDPOINT"):
-            kwargs.update(
-                aws_s3_endpoint=settings.RASTER_AWS_S3_ENDPOINT
-            ) 
+        try: 
+            kwargs.update(aws_s3_endpoint=settings.RASTER_AWS_S3_ENDPOINT)
+        except AttributeError:
+            pass
+             
 
         cache_key = hash(ensure_hashable(kwargs))
 
--- a/terracotta/drivers/geotiff_raster_store.py
+++ b/terracotta/drivers/geotiff_raster_store.py
@@ -130,6 +130,15 @@ class GeoTiffRasterStore(RasterStore):
         if tile_size is None:
             tile_size = settings.DEFAULT_TILE_SIZE
 
+        if settings.RASTER_AWS_ACCESS_KEY is not None:
+            self._RIO_ENV_OPTIONS = self._RIO_ENV_OPTIONS.update(
+                AWS_VIRTUAL_HOSTING=False,
+                AWS_HTTPS='NO',
+                AWS_S3_ENDPOIN=settings.RASTER_AWS_S3_ENDPOINT,
+                AWS_ACCESS_KEY_ID=settings.RASTER_AWS_ACCESS_KEY,
+                AWS_SECRET_ACCESS_KEY=settings.RASTER_AWS_SECRET_KEY
+            )
+
         kwargs = dict(
             path=path,
             tile_bounds=tile_bounds,
--- a/terracotta/drivers/geotiff_raster_store.py
+++ b/terracotta/drivers/geotiff_raster_store.py
@@ -140,10 +140,8 @@ class GeoTiffRasterStore(RasterStore):
             rio_env_options=self._RIO_ENV_OPTIONS
         )
         
-        try: 
+        if settings.RASTER_AWS_S3_ENDPOINT is not None:
             kwargs.update(aws_s3_endpoint=settings.RASTER_AWS_S3_ENDPOINT)
-        except AttributeError:
-            pass
              
 
         cache_key = hash(ensure_hashable(kwargs))
--- a/terracotta/drivers/geotiff_raster_store.py
+++ b/terracotta/drivers/geotiff_raster_store.py
@@ -74,7 +74,6 @@ def ensure_hashable(val: Any) -> Any:
 
     return val
 
-
 class GeoTiffRasterStore(RasterStore):
     """Raster store that operates on GeoTiff raster files from disk.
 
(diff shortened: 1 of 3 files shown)


MatthewTe:
Not a huge fan of how the AWS config variables get passed into the `get_raster_tile` method - would appreciate any suggestions to clean this up

terracotta/raster.py
```diff
@@ -316,13 +316,38 @@ def get_raster_tile(
...
+        rio_env_options.update(
+            AWS_VIRTUAL_HOSTING=False,
+            AWS_HTTPS='NO',
+            GDAL_DISABLE_READDIR_ON_OPEN='YES',
+            session=aws_session
+        )
+
+        del rio_env_options['AWS_ACCESS_KEY']
+        del rio_env_options['AWS_SECRET_KEY']
+        del rio_env_options['AWS_S3_ENDPOINT']
```
dionhaefner, line 341: So this will spawn a separate session for every tile, which to me seems like a performance killer. To do this properly, we need some machinery to re-use a session for each worker...

dionhaefner: Maybe a simple memoized function `get_aws_session` would do?

dionhaefner: You can also check for `settings.RASTER_AWS_S3_ENDPOINT` directly without pushing / popping stuff from `rio_env_options`.

MatthewTe: > Maybe a simple memoized function `get_aws_session` would do?

I've made a very basic `get_aws_session` method that uses the python functools lru cache - I didn't realize that the `AWSSession` class `TerracottaSettings`  was not hashable/pickleable so everything needs to be accessed directly in the `get_raster_tile`.

I've made the `get_raster_tile` use the `AWSSession` based on the `settings.RASTER_AWS_S3_ENDPOINT` passed in as a kwarg.

MatthewTe:
`get_aws_session` currently takes in no parameters so technically its not memoized and won't cache invalidate if the settings change. I could make it accept the `aws_s3_endpoint`  as an input parameter to force memoization tied to the settings object.....

terracotta/drivers/geotiff_raster_store.py
```diff
@@ -138,8 +137,14 @@ def get_raster_tile(
...
             resampling_method=settings.RESAMPLING_METHOD,
             target_crs=self._TARGET_CRS,
-            rio_env_options=self._RIO_ENV_OPTIONS,
+            rio_env_options=self._RIO_ENV_OPTIONS
         )
+        
+        try: 
+            kwargs.update(aws_s3_endpoint=settings.RASTER_AWS_S3_ENDPOINT)
+        except AttributeError:
+            pass
```
dionhaefner, line 146: This should never trigger. I think what you want is this:

```suggestion
        if settings.RASTER_AWS_S3_ENDPOINT is not None:
            kwargs.update(aws_s3_endpoint=settings.RASTER_AWS_S3_ENDPOINT)
```

dionhaefner:
> get_aws_session currently takes in no parameters so technically its not memoized and won't cache invalidate if the settings change. I could make it accept the aws_s3_endpoint as an input parameter to force memoization tied to the settings object.....

Or just pass the entire `settings` object? Whichever is easiest really, settings should only change rarely but it's good to be defensive to prevent hard-to-debug issues from creeping into the codebase.

terracotta/drivers/geotiff_raster_store.py
MatthewTe: oh wait your right - forgot all of the variables get assigned a default value

MatthewTe:
> > get_aws_session currently takes in no parameters so technically its not memoized and won't cache invalidate if the settings change. I could make it accept the aws_s3_endpoint as an input parameter to force memoization tied to the settings object.....
> 
> Or just pass the entire `settings` object? Whichever is easiest really, settings should only change rarely but it's good to be defensive to prevent hard-to-debug issues from creeping into the codebase.

I wold just pass  in `RASTER_AWS_S3_ENDPOINT` as an unused input arg to make it memoize to the S3 bucket url so we don't have to call the `get_settings` in `get_raster_tile` method unless a user updates the `RASTER_AWS_S3_ENDPOINT` and avoid an additional `get_settings` call. 

Decided  I should just pass in `settings` to avoid the weird case of the access/secret key changing but the endpoint staying the same which would result in unexpected behaviour (even if it doesn't change very often).

terracotta/raster.py
```diff
@@ -20,11 +22,29 @@
...
-from terracotta import exceptions
+from terracotta import exceptions, get_settings
 from terracotta.profile import trace
 
 logger = logging.getLogger(__name__)
 
+@functools.cache
+def get_aws_session(RASTER_AWS_S3_ENDPOINT: str) -> Union[AWSSession, None]:
+
+    settings = get_settings()
```
dionhaefner, line 33: Seems like we could avoid most caching issues if we just add all AWS settings as args here, and call `get_settings` in `get_raster_tile`? 

Alternatively we could amend the signature of `get_raster_tile` to accept `aws_s3_config: dict[str, str]` and pass that verbatim as kwargs to `get_aws_session`, putting the burden to assemble it on the driver (which is probably where this logic belongs anyway).

dionhaefner:
And please fix those `pre-commit` issues – I suggest you just run `pip install pre-commit && pre-commit install` on your end to catch them automatically.

terracotta/raster.py
MatthewTe: Yep thats probably for the best, moved the logic out of the `get_raster_tile` method and into the driver. Simplified it a bit that way

dionhaefner (commented):
Perfect, thank you! Could you please give this a final spin as to whether it works for your use case?

MatthewTe:
> Perfect, thank you! Could you please give this a final spin as to whether it works for your use case?

Yep! Just did a quick check and it serves my purpose - serving geotiffs through S3 authentication. Theoretically another feature to add could be to extend this authentication to the `sqlite_remote_meta_store` as right now you wouldn't be able to upload/download the sqlite database that requires authentication. 

the sqlite driver already uses boto3 though so it might already support reading from S3 buckets w/ AWS auth via the same environment variables but its not a feature required for my use case. Perhaps something for me to look into later :)

dionhaefner (approved):
Perfect, thanks a lot!

Issue #360, opened 2025-04-11
OverflowError: math range error on too high `z`


jkittner:
It is quite easy to trigger internal server errors by setting unrealistically high z-values. This can happen quickly by confusing `z` with `x` or `y` when setting up a TMS e.g. in QGIS or similar.
As far as I can no input validation is done and the bug is upstream in mercantile.
I will try to fix it upstream so a proper error is raised that we can except here and then treat the tile as non-existing.

I will maybe also send a simple alternative here, since the upstream does not seem to be very active.

I created: https://github.com/mapbox/mercantile/pull/153

PR #361, opened 2025-04-11
avoid OverflowError by limiting z

jkittner:
this would previously cause 500s when accidentally swapping z with x or y

addresses #360

### DIFF
--- a/terracotta/xyz.py
+++ b/terracotta/xyz.py
@@ -3,12 +3,16 @@
 Utilities to work with XYZ Mercator tiles.
 """
 
+import math
+import sys
 from typing import Optional, Sequence, Union, Mapping, Tuple, Any
 
 import mercantile
 
 from terracotta import exceptions
 from terracotta.drivers.terracotta_driver import TerracottaDriver
+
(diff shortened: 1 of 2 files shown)


dionhaefner:
Thanks @jkittner. Where does the value 30 come from?

jkittner:
I thought that was a sane default for TMSs. I could do some more research if this is specced out somewhere maybe.

We could also go with the theoretical maximum of `int(math.log2(sys.float_info.max)) - 1`.

Tbf it's not really an integer overflow since `pow` alway returns float. I should probably correct that.

dionhaefner:
I think for now a good value would be the max we can get without triggering that particular overflow error.

jkittner:
Okay, I changed it to the maximum value. That's definitely 100 % backwards compatible.

Seems like the gha pre-commit action cache is being deprecated. Maybe worth switching to pre-commit.ci: https://pre-commit.ci/ ?

dionhaefner (approved):
Perfect, thanks! I'll fix the CI failures in a separate PR.

PR #362, opened 2025-04-15
Fixing zoom to file in App Preview

zufryy:
This would fix the zoom to clicked file function in the App Preview


dionhaefner:
Hmm looks like the client build action isn't allowed to push to your fork. @zufryy could you try to run the rebuild manually please, and commit the result?

PR #365, opened 2025-04-16
fix many deprecation warnings

jkittner:
the warnings came mainly from marshmallow and cpython itself

- you cannot provide metadata via kwargs to scheams in marshmallow any longer
- validator must not return bools any longer - just raise an exception upon failure
- some minor cpython stuff e.g. in the ast module

In total that should have removed a couple hundred warnings - and I enabled them again, so we can actually see if something is deprecated.

There is a warning left with multiprocessing I cannot find the root cause of.

### DIFF
--- a/terracotta/config.py
+++ b/terracotta/config.py
@@ -116,8 +116,9 @@ DEPRECATION_MAP: Dict[str, str] = {
 }
 
 
-def _is_writable(path: str) -> bool:
-    return os.access(os.path.dirname(path) or os.getcwd(), os.W_OK)
+def _is_writable(path: str) -> None:
+    if not os.access(os.path.dirname(path) or os.getcwd(), os.W_OK):
+        raise ValidationError(f"Path {path} is not writable.")
 
 
 class SettingSchema(Schema):
--- a/terracotta/drivers/geotiff_raster_store.py
+++ b/terracotta/drivers/geotiff_raster_store.py
@@ -58,7 +58,7 @@ def submit_to_executor(task: Callable[..., Any]) -> Future:
         future = _executor.submit(task)
     except BrokenProcessPool:
         # re-create executor and try again
-        logger.warn("Re-creating broken process pool")
+        logger.warning("Re-creating broken process pool")
         _executor = create_executor()
         future = _executor.submit(task)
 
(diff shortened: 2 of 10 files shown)


setup.cfg
```diff
@@ -10,14 +10,8 @@ addopts =
...
   -v
 filterwarnings =
-  error
-  # ignored by default
-  ignore::ImportWarning
-  ignore::DeprecationWarning
-  ignore::PendingDeprecationWarning
-  # raised by Cython, usually harmless
-  ignore:numpy.dtype size changed:RuntimeWarning
-  ignore:numpy.ufunc size changed:RuntimeWarning
```
dionhaefner, line 20: Can we roll this back please? The first line is what made CI failing when a warning is raised, and the rest of them makes it so we don't fail on less severe warnings.

jkittner:
I restored the ignores to the previous state:

we had 378 warnings on 3.12 before this patch. We are now down to 224 and with #364 we will be down to 216.

What remains are these warnings:

```console
tests/drivers/test_raster_drivers.py: 14 warnings
tests/scripts/test_optimize_rasters.py: 182 warnings
tests/server/test_cors.py: 3 warnings
  /usr/lib/python3.12/multiprocessing/popen_fork.py:66: DeprecationWarning: This process (pid=152991) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()

tests/drivers/test_sqlite_remote.py: 17 warnings
  /home/kittnjdr/workspace/terracotta/.tox/py312/lib/python3.12/site-packages/botocore/auth.py:425: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    datetime_now = datetime.datetime.utcnow()
```

The boto one is clearly not our fault and needs fixing upstream - the former ones likely are and I believe are due to `ThreadPoolExecutor`?

dionhaefner:
As long as tests are passing I'm not too worried about deprecation warnings. The majority of them will be fixed by other people (3rd party maintainers) when they become a problem. For the remainder, we'll deal with them when they actually break something; at that point there are usually enough guides on how to update that this transition is much easier to deal with.

dionhaefner (approved):
Good fixes, thanks for the cleanup!

jkittner:
btw. marshmallow 4.0.0 release 10 hours ago (https://pypi.org/project/marshmallow/4.0.0/) so version 0.9.0 currently does not install cleanly w/o specifying `marshmallow<4.0.0`. This patch/the main branch should have already fixed that.

PR #366, opened 2025-04-19
Fixing zoom to file in App Preview - Rebuild

zufryy:
This would fix the zoom to clicked file function in the App Preview - Rebuild Manually


dionhaefner:
Awesome, thanks @zufryy!

PR #367, opened 2025-06-18
add order_by and ascendig query parameters to datasets endpoint

jkittner:
I have the use case where I want to query the tms for the latest available dataset so my front end can decide what tiles to load.

The current approach is to just filter by year and query a lot (e.g. 5000), then sort them in the front end / svelte app and decide what's the latest.

With this you could simply set the `limit` to 1 and specify the keys you want to order by and decide via `ascending`whether it should be ascending or descending meaning get the latest or oldest data.

This is just a quick draft to get a feeling on how complex this change would be. Just let me know if you would be interested in such feature!

...The marshmallow validation is bit finicky with lists. In http it would be also valid to simply specify a parameter multiple times, but that does not work. It only works with the literal list. This would need a closer look.


dionhaefner:
Sounds like a useful feature. Lmk if I can help.

jkittner:
Mhm, the test failures seem unrelated?!

```
E        ACTUAL: array(3608.971706)
E        DESIRED: array(3608.9712, dtype=float32)
```

Looks like a floating point glitch? Maybe the tolerances need adjusting? Or it used to be float16 and is now float32?

dionhaefner:
That's weird. My money would be on a dependency (rasterio?) that introduced custom code paths for something, or changed an internal data type :confused:

dionhaefner:
Oh I see, this only happens when using crick. In that case some deviation is expected and the suggested fix should be fine. Thanks for investigating!

jkittner:
I just ran I to the issue I made the workaround for back then.

> The current approach is to just filter by year and query a lot (e.g. 5000), then sort them in the front end / svelte app and decide what's the latest.

So I was wondering if we could give this another try?

I would rebase and check the tests again.

dionhaefner:
Sounds good, if the tests are green this should be gtg from my side.

setup.cfg
```diff
@@ -21,6 +21,8 @@ filterwarnings =
   # sometimes, dependencies leak resources
   ignore:unclosed database in <sqlite3.Connection:ResourceWarning
   ignore:.*socket\.socket.*
+  # rasterio deprecated antimeridian_cutting parameter but still uses it internally
+  ignore:The antimeridian_cutting and antimeridian_offset parameters:rasterio.errors.RasterioDeprecationWarning
```
jkittner, line 27: https://github.com/rasterio/rasterio/blob/d1a99cca77bd88997552884d706ca8b668546efa/rasterio/features.py#L761-L764

That got me thinking a little where this comes from. But turns out rasterio deprecated some function arguments but still uses them internally which triggers this warning.

the warning in question: https://github.com/rasterio/rasterio/blob/d1a99cca77bd88997552884d706ca8b668546efa/rasterio/warp.py#L97-L102

jkittner:
Python 3.9 is end of life - that's why it fails (boto3). Do we want to drop it? I would do that in a separate PR if you agree.

dionhaefner:
Sure, should be safe to bump the supported range to `[3.10, 3.14]`. Thanks for looking into it.

dionhaefner (approved):
gtg from my side, @jkittner is this ready?

jkittner:
I see you merged main already - so yes, good to go!

dionhaefner:
Thanks!

jkittner:
Great thank you! Any chance we could make a release with this?

dionhaefner:
[Beat you to it](https://github.com/DHI/terracotta/releases/tag/v0.10.0) 😏

PR #368, opened 2025-07-31
fix pillow deprecation warning for mode argument in fromarray

jkittner:
This parameter is scheduled for removal: https://pillow.readthedocs.io/en/latest/deprecations.html#image-fromarray-mode-parameter

They state:

> The mode can be automatically determined from the object’s shape and type instead.

Simply removing it does the trick. This is covered by tests which still pass.


dionhaefner (approved):
lgtm, thanks!

jkittner:
It's this I think

https://github.com/DHI/terracotta/pull/367#issuecomment-3004215755

PR #370, opened 2025-10-31
Fix docker build github action

atanas-balevsky:
This PR would fix the failing docker build push github action
- upgrades the base image from `bullseye` to `bookworm`
- installs required packages setuptools and wheel before running the `setup.py`


atanas-balevsky:
> Thanks @atanas-balevsky! Was this tested somewhere? Why were the Docker builds broken in the first place, and how did the Debian upgrade fix it?

from the failing logs you can see the build failing for missing wheels and setuptools. 
after i got that, the next build error was for an old version of gdal. bullseye is a bit oldish debian, so i switched to the newer stable. 
i'm not sure if you upgraded gdal in terracotta at some point or when the docker build started failing, but the latest version in hub.docker.com is a bit old

PR #371, opened 2026-01-07
drop python3.9

jkittner:
dropping python3.9 as discussed in #367 - I also cherry-picked the rasterio deprecation warning commit since this would fail otherwise.


jkittner:
always something...

```
E                   pytest.PytestUnraisableExceptionWarning: Exception ignored while calling deallocator <function _TemporaryFileCloser.__del__ at 0x7fc5b8d98930>: None
```

...will look into this later

dionhaefner (approved):
Awesome, thanks!

PR #372, opened 2026-01-12
add missing migrations

jkittner:
I just ran my deploy, but we forgot to add the migrations for the (major) version...

The startup fails due to a version mismatch, but we cannot perform a migration because it is missing. Even though it's a noop migration we need it.

I changed the tests to **NOT** skip when the next, upcoming migration is missing. This should prevent that from happening again.

Part of the release should be adding a new (upcoming) migration. So releasing 0.11 should instantly add the 0.12 migration - otherwise the tests will fail.

This is not perfect, but since we need to upgrade the db version, it's not a true noop migration, so just skipping is not a great idea...


jkittner:
Mhm. Thinking about this again. The 0.11. file must be added AFTER the release was made. But that will cause the tests to fail with the release.

What do you think about keeping it simple and just running 

```sql
UPDATE terracotta SET version='0.10.0'
```

if there are no migrations?

dionhaefner:
I like what you have implemented. I can see how it's not perfect but failing that test on release versions shouldn't be too consequential.

There is a lot more we could do but given that we release only a few times per year this is probably low ROI to fix. Merging for now.

